PR #24, opened 2021-04-28
Override TSDX; Upgrade storybook; Fix build errors

WilliamKelley:
While TSDX is great, it seems to be under long-term maintenance. A once active milestone history has been stale for over a year. The main maintainer seems to be taking an extended leave from open-source due to "toxicity/abuse". Notable blocked upgrades include TS v3.8+, TS-Jest v24+, Prettier v2+, ESLint v7+, TS-ESLint v3.8+.

After reading through the various backlogs and history of TSDX, I honestly think the maintainer is justified in taking a break, and that generally TSDX is a very complicated project, focused on making a clear upgrade path without major breaking changes -- a very tough feat when it combines so many technologies with various versioning habits.

However, these older versions do limit my dev experience and ease -- and I imagine others.
- Prettier v1 conflicts with `import type` or `export type` but Webpack / the build process warns for every TS type/interface not exported with `export type`.
- TS v3.8 feature-set and documentation takes significantly more effort to find and validate as time goes on and v4 becomes standard discussion & documentation for over a year.
- TS v4+ includes many nice features relevant for this project, such as template types.

The endorsed workaround is to specify the packages in the `"resolutions": {...}` field of `package.json`. This is apparently a feature of Yarn (not NPM, yet) which has the ability to override the internal dependencies of TS. (It should be noted that TSDX does not use a "peer-dependency" model where we could use whatever version of these packages as the peer through a simple install. This is the crux of the issue)
- The first commit of this PR accomplishes that with `yarn install` .

At this point, I should say that I believe TSDX combines the use of Yarn and NPM under the hood, and has some mechanism for us to be able to use the feature-set of both. The standard installs of each (`yarn` and `npm install`) seem to also run `tsdx build`. It seems to be necessary that every dev will need to run `yarn install` for the overrides, and I think `npm install` for the storybook upgrades i'll talk about later. This doesn't make the most sense to me, so maybe I'm wrong in this regard. (But it does answer another question brought up in Spark: we should not delete `yarn.lock` it's required by TSDX as far as I can tell)

The second commit fixes many build warnings (that will only increase with every component we add) by using `export type ...`.

The third commit runs Prettier v2.

The fourth commit runs `npx sb@latest upgrade` -- the advisable upgrade mechanism for Storybook.
- I think this is completely separate from TSDX, but I thought it was fine to include.

The fifth commit runs `yarn upgrade`
- Again separate from my TSDX concerns, but is a good chore and nothing broke?

The sixth commit fixes the `dist/` build error (blocking) from Webpack by switching to using just imports from `'../../src'` in our Stories -- I'm not sure if this is bad practice or not, if anyone knows please enlighten me. Half of our stories did this, half did not. It seems that with Storybook v6.2+, the build process won't let us import from `'../../dist'` -- I'm not sure why, but my guess is it just doesn't make the files available in time. 

The seventh/last commit runs `npm audit fix` for 18 moderate warnings (of 21). Upgraded only internal dependencies. Confirmed nothing broke -- including the near-indeterminate behavior that the last audit fix produced, that's not happening now.

Overall, this PR is large and covers probably too much ground, but there's a common thread and I think it's healthy for the project. Let me know your thoughts.


WilliamKelley:
@haase1020 

> 
> 
> Do we need to use TSDX? Would it be possible to remove that package entirely?

You can see everything TSDX does for us at it's [website here](https://tsdx.io/). You should create an Issue to discuss removing it entirely. There will be a lot to discuss, and it's not the aim of this PR.



> 
> 
> In order to move forward with development in spark, it would be very nice to have this PR approved and merged, or to come up with another solution. Currently, I am running into problems trying to get storybook started up per the `README` instructions.

Would be great if you could create an Issue detailing the problems your running into. Then we can make changes accordingly

pinindajin (approved):

WilliamKelley:
I've merged main to resolve conflicts. The recent Button fix pr merge deleted the yarn.lock, but I restored it since we're using yarn's resolutions feature that npm cannot match.

To that end, I did some more research and I'm going to advise deleting `package-lock.json` in favor of `yarn.lock` and moving towards exclusively using `yarn` instead of `npm`. As mentioned, this is because we want to ensure the overriding of preset package versions in tsdx, and only `yarn` can do that. We can completely replace usage of `npm` because they do the same thing otherwise (both use package.json)

I've update the README appropriately.

---

ALSO, I discovered the source of the build errors preventing successful compiling on the previous `npm run storybook` / `yarn run storybook` invocation. It was set to run `tsdx watch` and `start-storybook` _concurrently_, however storybook would start faster the tsdx could build! Since this appeared after upgrading storybook, I'm going to infer that their package either evaluates imports earlier or is faster in general than in previous versions.

In fact, that concurrent call was a deviation from the [official guide here](https://github.com/formium/tsdx/blob/master/templates/react-with-storybook/README.md) (I created a test react-with-storybook project with tsdx to confirm). The consequence is that we should stay true to the official guide, and require devs to run `yarn start` (tsdx watch) in one terminal and `yarn run storybook` in another, after a successful first build from the watch.

I've restored all the `../../dist` import references in the stories, so they used the exact production output, with no build errors.

Again, I've updated the README appropriately.

PR #47, opened 2021-06-01
Chore: format all

WilliamKelley:
Effect of `nx format:write --all`.

I will introduce a second PR to add a pre-commit hook to do this automatically.


WilliamKelley:
Merged without review because it's just a simple chore

PR #48, opened 2021-06-01
Chore: lint all

WilliamKelley:
Depends on #47

Fixes `"lint"` script in `package.json` by replacing `nx lint` with `nx run-many --all --target=lint --fix`. The former fails because a project name is required, the latter lets NX find all projects without individual specification.

I fixed all the existing lint errors / warnings.


WilliamKelley:
Merged without review because it's just a simple chore

PR #49, opened 2021-06-01
Repo: lint & format pre-commit & -push

WilliamKelley:
[Adds new husky hooks](https://typicode.github.io/husky/#/) to run the `"lint"` and `"format"` scripts pre-commit and pre-push. As usual, each op will fail if the husky hooks fail.


WilliamKelley:
Going forward with merging without review since it's a sensible default

PR #50, opened 2021-06-01
Add: nx semver plugin

### LINKED ISSUE #44: SemVer Changelog: auto-gen by adhering to "Conventional Commits" (plugin)
Spark rev has a stated intention to adhere to SemVer, however our alpha dev has not quite lived up to that so far. And, there is no formal Changelog created or updated for each v-bump.

Moving forward, now that multiple Prenda projects are beginning to rely on Spark, keeping an accurate changelog will be critical to Spark's good future. I foresee breaking changes before we can fill out dev for the rest of the Figma designs.

I propose we address these intertwined issues with an NX plugin, [`@jscutlery/semver`: "Nx plugin for versioning using SemVer and CHANGELOG generation powered by Conventional Commits."](https://github.com/jscutlery/semver).

This plugin will require that Spark dev's adhere to the [Conventional Commits guidelines](https://www.conventionalcommits.org/en/v1.0.0/) in order for it to do its job right. I think the _examples_ and _spec_ sections do a great job at explaining the requirements of conv. commits and should be straightforward for all Spark devs to adhere to. The Why/FAQ sections have as good of answers to why we should adopt this plugin as I could write. Some highlights:
- [Fixing commit message mistakes](https://www.conventionalcommits.org/en/v1.0.0/#what-do-i-do-if-i-accidentally-use-the-wrong-commit-type)
- [Moving responsibility from every dev to lead dev / PR merger by adopting a squash flow](https://www.conventionalcommits.org/en/v1.0.0/#do-all-my-contributors-need-to-use-the-conventional-commits-specification)
    - I think this is a viable option for us.

There are also templates and IDE plugins for Conventional Commits on the website that can help us.

WilliamKelley:
Closes #44 

Adds an NX plugin to aid us in correct Semantic Versioning for our projects:
- specifies a commit convention to track new features, changes (including breaking), and other: [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/).
- provides a release mechanism: `nx run <project>:version` (e.g. `nx run spark:version`)
  - goes through commit history to determine correct version increment
  - auto-tags commit
  - auto-generates a CHANGELOG.md that we can publish / make available to consumers

I had to tag the last commit on `main` with the 0.7.1 version in order for this plugin to not revert us to 0.0.0 . Spark won't have a change-log for anything below 0.7.1 but someone can go through and add it manually if they need to I suppose.

The most important part of this is that everybody reads through [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) and abides by this syntax for at least all commits that matter.

As noted in the [source issue](#44), we could setup a squash-and-merge PR procedure where the lead reviewer can aggregate or correct commit history, but I think that's a last resort.


shadoath (approved):
Having conventional commits is a good idea for Spark

seigler (approved):
:ship:

PR #71, opened 2021-06-18
feat: theme overrides for Card, CardContent, CardActions

seigler:
image

Simple Card component supporting a heading image.

Also in this PR:
* pre-commit and pre-push hooks now only lint affected code
* Material UI shadows in the Prenda theme have been lightened 50%


seigler:
Please merge #70 before this so the release tags are in the right order

libs/spark/src/theme/theme.ts
```diff
@@ -107,6 +108,33 @@ export const prendaTheme: ThemeOptions = {
...
+    createShadow(0, 8, 9, -5, 0, 15, 22, 2, 0, 6, 28, 5),
+    createShadow(0, 8, 10, -5, 0, 16, 24, 2, 0, 6, 30, 5),
+    createShadow(0, 8, 11, -5, 0, 17, 26, 2, 0, 6, 32, 5),
+    createShadow(0, 9, 11, -5, 0, 18, 28, 2, 0, 7, 34, 6),
+    createShadow(0, 9, 12, -6, 0, 19, 29, 2, 0, 7, 36, 6),
+    createShadow(0, 10, 13, -6, 0, 20, 31, 3, 0, 8, 38, 7),
+    createShadow(0, 10, 13, -6, 0, 21, 33, 3, 0, 8, 40, 7),
+    createShadow(0, 10, 14, -6, 0, 22, 35, 3, 0, 8, 42, 7),
+    createShadow(0, 11, 14, -7, 0, 23, 36, 3, 0, 9, 44, 8),
+    createShadow(0, 11, 15, -7, 0, 24, 38, 3, 0, 9, 46, 8),
```
shadoath, line 136: Can you explain this to me?

seigler: This is the same code Google uses to generate the different levels of shadow corresponding to the `elevation` prop on most of their components. But I changed the opacity constants in `shadows.ts` to be 50% lighter to more closely match Spencer's designs.

libs/spark/src/Card.tsx
```diff
@@ -0,0 +1,40 @@
+import styled from 'styled-components';
```
shadoath, line 1: Can be removed.

libs/spark/src/theme/theme.ts
seigler: The numbers correspond to pixel sizes for shadow location, offset, and blur.

shadoath (commented):
Minor change. 
Didn't know about https://placekitten.com fun service for sure.

libs/spark/src/theme/theme.ts
shadoath: Ahhh, then the array index correlates to the `elevation` wanted.

WilliamKelley (changes requested):
This PR is missing following the [Contributor](https://github.com/prenda-school/prenda-spark/blob/main/CONTRIBUTING.md#new) guideline on new components and on implementation work. Doing so could alleviate a lot of my review. The customization of theme shadows and the correcting of the lint command should have been separate PR's. Those could have been approved/merged no problem, without being held up by Card review.

(Note: I don't mean to be over-bearing / pedantic here; I push because I think each of the guidelines has value. If it's ill formed / too strict, lets work together to improve it.)

This implementation could be greatly simplified. This biggest issue is forcing the first child of `Card` to be `CardContent`, which is unnecessary / harmful to use-cases. MUI has [several components related to Cards](https://material-ui.com/components/cards/#api). Spark would be better to customize each of these individually, and leave assembly up to the consumer -- this is standard re-usability under React. Spark will undoubtedly need Cards with varying layouts (text, image organizations, and other) in the future, and assuming one composition will force a rewrite / breaking change in the future.

Action items
- Adhere to guidelines
- Customize one component per file
- Customize each card component needed

libs/spark/src/theme/theme.ts
WilliamKelley: This is very nice to have customized!

Also small note: Material-UI /= Google. The connection between them is that Material-UI creates libs that implement Google's published 'Material Design', but they're very much different companies.

WilliamKelley:
I don't think this PR should version itself, it poses problems for changelog generation and review, see #72

libs/spark/src/Card.tsx
```diff
@@ -0,0 +1,38 @@
+import { withStyles } from '@material-ui/core/styles';
+import MUICard from '@material-ui/core/Card';
```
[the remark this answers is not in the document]
seigler: Thanks for this comment. In this instance it has become moot

libs/spark/src/theme/shadows.ts
```diff
@@ -0,0 +1,11 @@
+const shadowKeyUmbraOpacity = 0.1;
+const shadowKeyPenumbraOpacity = 0.07;
+const shadowAmbientShadowOpacity = 0.06;
+
+export function createShadow(...px) {
```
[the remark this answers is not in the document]
seigler: This was an excellent suggestion, which I've included.

libs/spark/stories/card.stories.tsx
```diff
@@ -0,0 +1,25 @@
...
+  title: 'prenda-spark/Card',
+  component: Card,
+  argTypes: {},
+  args: {},
+} as Meta;
+
+export const BasicCard = () => (
+  <Card>
+    <h2>Card Header</h2>
+    <p>Card body</p>
```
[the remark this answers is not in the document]
seigler: Typography needs its own theme PR to match Spencer's designs.

seigler:
I rolled back the versioning and release tag, and moved the changes to Card and CardContent defaults into the theme.

libs/spark/stories/card.stories.tsx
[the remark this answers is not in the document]
WilliamKelley: we have a typography component :)

libs/spark/stories/card.stories.tsx
```diff
@@ -0,0 +1,51 @@
...
+const MyHeadingImg = styled.img`
+  width: 100%;
+  max-height: 300px;
+  object-fit: cover;
+  object-position: 50% 10%;
+  vertical-align: bottom;
+`;
+
+export const CardWithImage = () => (
+  <Card>
```
WilliamKelley, line 35: let's one-off customize this and add a `max-width: 580px` so it's not huge in the story. I chose that value from the figma page example.

libs/spark/stories/card.stories.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+export const BasicCard = () => (
+  <Card>
+    <CardContent>
+      <Typography gutterBottom variant="h6" component="h2">
+        Card Title
+      </Typography>
+      <Typography variant="body1" component="p">
+        Card body
+      </Typography>
```
WilliamKelley, line 21: For each of these stories, I'd suggest copying the text/content in the figma pages. It doesn't matter technically, but just to validate the overall look with a bunch of text is nice.

WilliamKelley (commented):
Overrides are so nice. The stories can advantage Spark's Typography.

There are a few other elements showcased in the Card Figma page: 'action' Button in the Footer, Icon Button at the top right, Tag beside title. We don't have tags yet, but we have the other two. Do you want to add validation and stories for them in this PR or leave it as is?

libs/spark/stories/card.stories.tsx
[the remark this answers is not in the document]
seigler: :man_facepalming: yep, there it is. Nice

libs/spark/stories/card.stories.tsx
seigler: I like this suggestion.

libs/spark/stories/card.stories.tsx
seigler: I updated both Card examples to match their Figma counterparts.

libs/spark/stories/card.stories.tsx
```diff
@@ -13,13 +16,20 @@ export default {
...
   <Card>
     <CardContent>
-      <Typography gutterBottom variant="h6" component="h2">
-        Card Title
+      <Typography gutterBottom variant="heading-md" color="textOnLight">
+        Heading
       </Typography>
-      <Typography variant="body1" component="p">
-        Card body
+      <Typography variant="paragraph-lg" color="textOnLight">
```
WilliamKelley, line 22: Match figma
```suggestion
      <Typography variant="paragraph-xl" color="textOnLight">
```

libs/spark/stories/card.stories.tsx
```diff
@@ -31,20 +41,30 @@ const MyHeadingImg = styled.img`
...
     </CardMedia>
     <CardContent>
-      <Typography gutterBottom variant="h6" component="h2">
-        Card Title
+      <Typography gutterBottom variant="heading-md" color="textOnLight">
+        Title
       </Typography>
-      <Typography variant="body1" component="p">
-        Card body
+      <Typography variant="paragraph-lg" color="textOnLight">
```
WilliamKelley, line 65: Match figma
```suggestion
      <Typography variant="paragraph-xl" color="textOnLight">
```

libs/spark/stories/card.stories.tsx
```diff
@@ -31,20 +41,30 @@ const MyHeadingImg = styled.img`
...
         alt="Kitten placeholder"
       />
+      <MyFloatingFavButton variant="outlined" size="large">
+        <HeartIconDuotone />
+      </MyFloatingFavButton>
     </CardMedia>
     <CardContent>
-      <Typography gutterBottom variant="h6" component="h2">
-        Card Title
+      <Typography gutterBottom variant="heading-md" color="textOnLight">
```
WilliamKelley, line 62: Same as above
```suggestion
      <Typography variant="heading-md" color="textOnLight" style={{ marginBottom: '1rem' }}>
```

WilliamKelley (approved):
Small changes to exactly line up with Figma. Good to merge after they're addressed.

_Edit: actually, before merge, I'll go in and accept the new stories as "baseline" in Chromatic_

libs/spark/src/CardContent.tsx
```diff
@@ -0,0 +1,5 @@
+export const MuiCardContentStyleOverrides = {
+  root: {
+    padding: '24px 24px 0',
```
WilliamKelley, line 3: Can you check if this works & include if you think it's a good idea. Finding the right paradigm for padding when the children can vary is always hard.
```suggestion
    padding: '24px 24px 0',
    '&:last-child': {
      paddingBottom: 24;
    }
```

seigler: The last-child rule is already present from the Mui default styles, so this combination of rules already applies to MuiCardContent.

libs/spark/stories/card.stories.tsx
seigler: Might it be better to change the `marginBottom` rule for `Typography[gutter]`?

libs/spark/src/CardContent.tsx
WilliamKelley: Ah okay, great

libs/spark/stories/card.stories.tsx
WilliamKelley: Possibly, it's a bit of a toss up though because a 1rem gutter is desirable on a heading, but probably not on a paragraph or smaller text...

WilliamKelley: Let's change it one-off for now but keep it in mind. If this keeps popping up, we can set a better default.

WilliamKelley: Looking at how Material-UI does it for their docs, they set a top/bottom margin for header variants and avoid `gutterBottom` on them. Spark doesn't have such a specification, but filing a design request would be the way to go for that.

WilliamKelley:
Great work. I've accepted the stories as baselines in Chromatic, and changed, the title to a Conventional Commit.

I'll squash and merge this time to see if what I'm proposing in #72 works, and post as an example.

PR #79, opened 2021-06-23
fix: default MUI override of body font-size

### LINKED ISSUE #78: SparkThemeProvider is overriding body font size to 0.875rem
image

Working on implementing the new Navbar into Prenda World and noticed that when wrapping the Navbar with the SparkThemeProvider, the font size for the body is being overwritten. Adding an `!important` to the body in main.less in Prenda World fixes this, but this seems like a deeper issue. 

Any thoughts on what's happening here? @WilliamKelley

WilliamKelley:
Closes #78

### DIFF
--- a/libs/spark/src/theme/theme.ts
+++ b/libs/spark/src/theme/theme.ts
@@ -93,6 +93,12 @@ export const prendaTheme: ThemeOptions = {
   typography: {
     fontFamily: '"Nunito", Avenir, sans-serif',
     fontSize: 16,
+    body1: {
+      fontSize: 16,
+    },
+    body2: {
+      fontSize: 14,
+    },
   },
   overrides: {
     MuiCssBaseline: {
--- a/libs/spark/src/theme/theme.ts
+++ b/libs/spark/src/theme/theme.ts
@@ -92,7 +92,7 @@ declare module '@material-ui/core/styles/createPalette' {
 export const prendaTheme: ThemeOptions = {
   typography: {
     fontFamily: '"Nunito", Avenir, sans-serif',
-    fontSize: 1,
+    fontSize: 16,
   },
   overrides: {
     MuiCssBaseline: {
--- a/libs/spark/src/theme/theme.ts
+++ b/libs/spark/src/theme/theme.ts
@@ -103,6 +103,10 @@ export const prendaTheme: ThemeOptions = {
   overrides: {
     MuiCssBaseline: {
       '@global': {
+        body: {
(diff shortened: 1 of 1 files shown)


WilliamKelley:
I'd like to have #77 merged first, need another review tho

WilliamKelley:
Hm, this broke some styling, I need to do more work

libs/spark/src/theme/theme.ts
```diff
@@ -92,6 +92,7 @@ declare module '@material-ui/core/styles/createPalette' {
 export const prendaTheme: ThemeOptions = {
   typography: {
     fontFamily: '"Nunito", Avenir, sans-serif',
+    fontSize: 1,
```
shadoath, line 95: What is `1` referencing? Should this be `16` 
Or maybe better:
```
    htmlFontSize: 16,
``` 
or
```
    bodyFontSize: 16,
```

WilliamKelley: Turns out it's pixels. I thought it was `rem` since it only accepts a number not a string (TS complains), but turns out the engine converts the px value to rem behind the scenes. We won't want to override `htmlFontSize` because of the a11y detriment noted on your link, but 16 is the correct value for `fontSize`

slingner (approved):

WilliamKelley:
Looks a bunch of things get a bit bigger font size because of inheriting increase. I'll need to re-validate.

WilliamKelley:
@shadoath fixed regressions

WilliamKelley:
Added fix for additional issue noted in #78 about the CssBaseline defaults

WilliamKelley:
They weren't actually regressions. 
- Typography: Accessory element margins changed very slightly, throwing off diff
- Colors: text font size inherited from body, not Typography component
- Icons: again, font size inherited from body -- i actually noticed these demos / components are wrong. They should use <SvgIcon> not <svg>... task for later

Good to go / review

libs/spark/src/theme/theme.ts
```diff
@@ -92,10 +92,21 @@ declare module '@material-ui/core/styles/createPalette' {
...
+    body2: {
+      fontSize: 14,
+    },
   },
   overrides: {
     MuiCssBaseline: {
       '@global': {
+        body: {
+          fontFamily: '"Nunito", Avenir, sans-serif',
+          fontSize: 16,
```
shadoath, line 108: Ahh so it was needed here too!

shadoath (approved):

Issue #75, opened 2021-06-22
Advantage theme overrides


WilliamKelley:
PR #71 introduces a new pattern of specifying theme overrides for component styles and default props. These values will be applied by Mui's styling system to any `Card`, `CardContent`, etc. component under the `ThemeProvider`. Since the component isn't re-exported from Spark the consumer must import from `@material-ui/core`.

This approach is valuable but not without its caveats.

### Pros
- Simplifies our implementations
- Improves default styling / behavior of components with internal inheritance
  - See [this Button default prop](https://material-ui.com/customization/globals/#default-props) that disables the touch ripple across all components. This is because all components with that ripple use some form of `ButtonBase`, probably. The same pattern applies in terms of styling overrides.

### Cons
- Overrides do not cover all cases of customization: ~e.g. the `Radio` and `Checkbox` components both replicate the internal structure of their Mui counterparts, but with custom SVG's for their checked / unchecked state.~ We must export custom components from Spark with this.
- In v4 of Material-UI, custom variants and colors cannot be added at the theme level. 
  - This changes in v5, see #61.
- Consumer has to know which components to import from Mui vs Spark.

### When to use

This approach isn't appropriate in all cases. Even if we apply this everywhere we can, Spark will still export custom components, e.g. custom props like variant and color additions. Also consider `NavBar`; it's not appropriate to globally theme `AppBar` to match what we want it to look like for `NavBar`, because `AppBar` is meant as a more general component.

### Targets

My assessment is that these components are able to be fully transitioned:
- [x] FormHelperText -- embedded in Input file, should be extracted
- [x] InputLabel -- embedded in Input file, should be extracted
- [x] FormControlLabel -- embedded in Radio / Checkbox, should be extracted / generalized, see #66
- [x] Radio
- [x] Checkbox

These components could be partially transitioned -- all styles that aren't specific to custom variants, colors, state (I've tried to prefix these with `Spark...-` in the past)
- [ ] All -- all components have some base level styling

And these fully in v5:
- [x] Icon -- custom colors
- [ ] Typography -- many custom variants, colors
- [ ] Button -- custom variants
- [ ] IconButton -- custom variants
- [x] Input -- custom state (success)

#### Stakeholders

@seigler

WilliamKelley:
Started work on `FormControlLabel` extraction in `refactor/form-control-label` branch

WilliamKelley:
[...]

PR #80, opened 2021-06-28
feat(FormControlLabel): add global style

WilliamKelley:
part of #75 

Shouldn't cause any visible changes


seigler (approved):
This is such a clean approach; it's nice to use MUI as it was designed.

shadoath (approved):

PR #85, opened 2021-06-28
feat(Menu)!: remove outdated implementation

WilliamKelley:
The Menu implementation is wrong and outdated with how we're doing stuff now. It should be named SelectField anyways. It's confusing for consumers to see this Menu component that can't be used.

Now that we have a `Dropdown` component, only a future `Select` component will eventually replace the functionality this tried to give.


seigler (approved):
This component does not spark joy.

shadoath (approved):

Issue #73, opened 2021-06-21
Separate `@prenda/spark-icons`, revise generation method


WilliamKelley:
Given that #68 introduced ~1700 icon files, we need to be very specific and careful about how we're defining and exporting them.

### Problems

- The 1700 files take up ~2.41 MB of raw space, and even more on a user's disk depending on their disk file format (see below).
  - This enormous file count and overall size can bring bundlers to a literal crawl if the library isn't correctly tree-shakable / or icons aren't otherwise directly importable -- I have a feeling that with all the problems consumers have had, our library currently falls under this category.
- Current directory layout prevents consumer from specifying the same [import transform plugin rule](https://material-ui.com/guides/minimizing-bundle-size/#option-2) that MUI advises.
- Icon svg files are not optimized and contain unneeded accessory info.

![image](https://user-images.githubusercontent.com/25781782/122787046-f0a72780-d282-11eb-91e2-8324148507be.png)

### Solution

- Create new package `@prenda/spark-icons` -- overall this will mirror [`@material-ui/icons`](https://github.com/mui-org/material-ui/tree/latest/packages/material-ui-icons)
- Locate all icons in top-level of `src/`
  - Mimic [MUI's naming scheme](https://material-ui.com/components/icons/#usage)
```bash
src
├── Activity.tsx # Line variant, default (no suffix)
├── ActivityFilled.tsx # Filled variant
├── ActivityDuotone.tsx # Duotone variant
```
- After bulk export from Figma, create a script with [SVGO](https://github.com/svg/svgo) that's similar to [MUI's script](https://github.com/mui-org/material-ui/blob/latest/packages/material-ui-icons/builder.js) to rename and optimize the `path` tags of the SVG. 
  - The optimized output ideally is: 
    - 1 path tag per icon (duotone maybe requires 2)
    - one `d` attribute per tag (`fillRule`, `clipRule` might be unavoidable by just using SVGO though, that'd be fine).
    - completely get rid of `svg` tag definition, the `SvgIcon` component takes care of this for us.
  - File mimics [MUI's layout](https://github.com/mui-org/material-ui/blob/latest/packages/material-ui-icons/src/AcUnit.js)
    - Copy their [internal `createSvgIcon` util](https://github.com/mui-org/material-ui/blob/42954448866f204b0cf7e35127d5d0af9c24d6fb/packages/material-ui/src/utils/createSvgIcon.js#L7) to our codebase

WilliamKelley:
[...]

PR #94, opened 2021-06-29
feat(SvgIcon): add global styles and props

WilliamKelley:
part of #75 

This new customization will replace the current `Icon` -- will be removed in another PR (since that's just a custom SvgIcon. I implemented the `contrast` prop by changing the `opacity` CSS property instead of switching to the `...LowContrast` version of the text colors. This was to simplify implementation. From my assessment, with something so simple as icons, it produces the exact same result.

This is also a prerequisite for #73

### DIFF
--- a/libs/spark/src/Checkbox.tsx
+++ b/libs/spark/src/Checkbox.tsx
@@ -111,11 +111,13 @@ function SparkCheckboxIcon(props: {
       })}
     >
       <SparkCheckboxUncheckedIcon
+        color="inherit"
         fontSize={fontSize}
         viewBox="0 0 22 22"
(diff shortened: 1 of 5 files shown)


seigler (approved):
It's not super pretty to have to wrap the component just to have the style react to our contrast prop, but this works fine until v5 is available.

WilliamKelley:
> It's not super pretty to have to wrap the component just to have the style react to our contrast prop, but this works fine until v5 is available.

True.

Also i should not, this differs from the previous `Icon` implementation as `color='navy'` and `color='grey'` are now `color='primary'` and `color='secondary'`, respectively.

shadoath (approved):

PR #97, opened 2021-06-29
feat(Menu): add global styles and props

### LINKED ISSUE #96: Changed Menu elevation
Now that shadows have been replaced with elevations, Menu has an elevation of 3 instead of a shadow.

[Fimga](https://www.figma.com/file/y6Ca8BdJtQEd3Of1saRieF/Spark-Design-System-Web?node-id=1692%3A2900)

WilliamKelley:
part of #75 

includes #85

closes #96


shadoath (approved):
Not sure I really track this one. Looks like cleanup as we go to a simpler menu though.

seigler (approved):

seigler:
> Not sure I really track this one. Looks like cleanup as we go to a simpler menu though.

@shadoath this and similar PRs allow Spark users to use normal Material UI components instead of importing a wrapped component. Normal Mui components will have our styles and default props applied by the theme.

WilliamKelley:
@shadoath Seigler is correct. The `DropdownMenu` doesn't need specific styling, and the existing styling matches the desired global styling of `Menu`. So this extracts that to a global override.

Issue #103, opened 2021-06-30
Simplify Figma JS/CSS Color names


WilliamKelley:
Currently, the names of our color values are unnecessarily long.

@ssbittle Are the "JS" and "CSS" names of spark's palette listed in "Spark Design System - Base, Colors" important to designers? If not, I'd like to simplify them for our implementation purposes.

ssbittle:
Not important. We can change them to be more simple. No problem.

WilliamKelley:
Great, thank you. I'll make a PR to get feedback from other devs on my new names, then I'll go ahead change the Figma page when it's finalized.

WilliamKelley:
@ssbittle Could you grant me edit access to the "Base" figma doc temporarily so I can replace those names

ssbittle:
Access updated.

WilliamKelley:
Done.

@ssbittle We also have JS values for these rows. Some of which are not reflected in the panels to the right, but are usually the values referenced in designs. Is it acceptable to add on to them as well?

![image](https://user-images.githubusercontent.com/25781782/124007910-d9b2b480-d9a9-11eb-90c0-d34c34abe520.png)

ex: 

"#0A3872" as Label - Medium

to

"*Hex*: #0A4872
*JS*: palette.brand.blue" in Source Code Pro font

or if creating another panel is more appropriate

ssbittle:
@WilliamKelley That makes perfect sense. And I'm personally cool with that change.

WilliamKelley:
@ssbittle Didn't have enough space for the prefixes, but I added them in the best way i could think of. The issue will close once we get our code in line with the new names.

ssbittle:
@WilliamKelley Thanks for doing this.

Issue #67, opened 2021-06-11
Proposal to solve ThemeProvider issues


WilliamKelley:
Several of our apps have tried to integrate Spark so far, and almost all have experienced one issue or another with `SparkThemeProvider`, whether it be a conflicting existing theme, wrong injection order, etc.

My proposal is that we don't export `SparkThemeProvider` at all anymore. Instead, I propose we 
- export the result of `createMuiTheme` as `theme`
- instruct the user on the best tree structure (i.e. what's in `SparkThemeProvider.tsx` currently)
- move all of our theme values to a single parent property of theme (i.e. everything we currently set on `theme.xxx` we set on `theme.spark.xxx`). _Except_, values that we are confident are sensible defaults or changes from the the default Material-UI theme -- example: our `error`, `warning`, `success`, `info` values are pretty standard and acceptable among any application.

I believe that after the NX migration, Spark is set up to depend on `@material-ui/...` as a _peer_-dependency, not direct, so it is safe to assume that every spark consumer will have the components needed to construct the tree.

If the consumer has existing customizations made to the theme, we can offer a preservation by replicating [MUI's internal theme merging pattern](https://github.com/mui-org/material-ui/blob/d7c0ae5958a274c6d7043b3bda8291a1d93fd3c1/packages/material-ui-private-theming/src/ThemeProvider/ThemeProvider.js#L54-L62)

One consequence is we'd have to update most all of our current theme accesses -- should just be tedious though, not difficult.

Instead of moving a lot under the completely custom value (`theme.spark`) we could re-consider what values go where in a standard MUI theme as well, now that we have a better understanding of what the Spark Design system will look like, what it's conceptions of primary / secondary are, etc.

WilliamKelley:
Another option that could increase compatibility with existing projects is advantaging `createGenerateClassName`, [docs here](https://material-ui.com/styles/api/#creategenerateclassname-options-class-name-generator), where we provide `"Spark"` as the `seed`.

PR #104, opened 2021-06-30
feat(palette): initial implementation, simplified names

WilliamKelley:
Part of #103 

### Problem
The color value names in Figma are long and tailored for exporting as lone constants, not as part of a palette object. They also don't reflect how they'll be organized in the palette. Now that we're importing what's basically the `palette` object for use in our style overrides, this is a noticeable pain point.

### Address
Similar to how `shadows.ts` exports a single `shadows` object, I've created `palette.ts` to export a single `palette` object.

This PR just introduces the new format for dev feedback on names and organization within the object. My goals were
- integrate more closely with the [default theme](https://material-ui.com/customization/default-theme/?expand-path=$.palette)
- eliminate value repetition
- eliminate re-specification of default Mui values -- the `createMuiTheme` utility will merge our customizations with defaults.
- lift "tertiary" values up, eliminating `palette.tertiary`
- move text values under `palette.text`
- shorten long names
  - example for the above two bullets: `palette.background.lightContrastText` -> `palette.text.onLight`
- remove specifying `light` and `dark` where the design system doesn't make any specifications
  - Ideally we could supply colors that Mui components could use effectively by default, but when I investigated such use, most components don't actually use those values directly. Instead, they lighten / darken the `main` by varying constants. (see [Alert](https://github.com/mui-org/material-ui/blob/master/packages/material-ui-lab/src/Alert/Alert.js))

### What's next
A subsequent PR will accomplish replacing `colors...` with `palette.$`, and relevant usages across our styles. That will be a breaking change. 

_Edit:_ Some of the goals also will ease #67. But that's a side-effect

### DIFF
--- a/libs/spark/src/theme/palette.ts
+++ b/libs/spark/src/theme/palette.ts
@@ -1,0 +1,103 @@
+// Palette properties with repeatedly used values are extracted
+const red = {
+  1: '#f7d2da',
+  2: '#de5160',
+  3: '#b83345',
+  4: '#851b2c',
+  5: '#520a18',
+};
+const orange = {
+  1: '#ffb78f',
+  2: '#ff7f4d',
+  3: '#e0662d',
+  4: '#a34814',
+  5: '#5c2704',
+};
+const yellow = {
+  1: '#fae6b2',
+  2: '#f7d054',
+  3: '#dbb13d',
+  4: '#8f691d',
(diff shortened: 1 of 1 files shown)


seigler (approved):
I _really_ like this.

shadoath (approved):
YES!!!

PR #105, opened 2021-07-01
feat(palette): move greys under grey, remove whites, navy, neutrals

WilliamKelley:
Part of #103 

As I was writing the new "JS" names into the Figma page, the neutrals ended up overlapping with each other because the names were so long. It's arbitrary, but I saw another opportunity to shorten a commonly-used names, and I believe it makes sense.
- `palette.neutral.white` / `palette.background.white` -> `palette.common.white` (a Mui default -- we don't need to provide our own white)
- `palette.neutral.lighterGrey` -> `palette.grey.lighter`
- `palette.neutral.lightGrey` -> `palette.grey.light`
- `palette.neutral.mediumGrey` -> `palette.grey.medium`
- `palette.neutral.darkGrey` -> `palette.grey.dark`
- `palette.neutral.navy` -> eliminated; it's equivalent to `palette.blue[5]`
  - I think it's reasonable to require memorizing `navy === blue[5]` so that we can completely do away with the `neutral` property

I left `brand` alone because it has a unique `orange`. I'd be open to reducing that as well, but I'm guessing that's less palatable than what I've done so far.


seigler (approved):
Simple is beautiful

shadoath (approved):
Good that you added the notes for Navy

Issue #106, opened 2021-07-01
Lazy load icons for Story


WilliamKelley:
While it's useful for us to provide a list of all the icons and their corresponding JS names, listing them all in our Storybook is resulting in terrible performance: 23s build, 10s reload, and my hardware is pretty good.

We should remove the import of all of them until we can build a solution like [Material-UI has for Material Icons](https://material-ui.com/components/material-icons/).

PR #107, opened 2021-07-01
chore(storybook): remove icons for performance

WilliamKelley:
Part of #106 

I removed the general import `import * as Icons...`, so all that's left is the `ExampleIcon`. 

That improved performance back to 13s build, <1s reload.


shadoath (approved):
Dang, that's impressive change in reload times.

seigler (approved):
Nice. This also means less refactoring once we move the icons into their own library.

PR #108, opened 2021-07-01
feat(theme)!: remove colors in favor of palette

### LINKED ISSUE #103: Simplify Figma JS/CSS Color names

WilliamKelley:
Implements "What's next" on #104 , Closes #103

Notably:
- removes re-exports of color values => internal consumers of `palette` must import directly from `src/theme/palette` which is desired in the primary case of component style overrides to avoid a circular dependency with `theme` creation.
- adds usage of `createPalette` so that we can use the full theme palette without having to generate a theme
  - as mentioned previously, using the theme isn't possible for style overrides because its a circular dependency

This PR also adds the necessary TS interfaces and module augmentations. I'll note a couple things that resulted from this work:
- Using `createPalette` greatly reduces the custom(ized) interfaces and augmentations I needed to declare.
- Within the module augmentations, they don't need to extend the original interfaces, doing so is a "recursive" reference error, because it's like you're going into that module and just specifying more properties.
  - Mirrors naming pattern of Mui internal
  - Augmenting `Palette` ensures that all consumer usages of `Theme` will be correctly typed.
  - Augmenting `PaletteOptions` ensures that `createPalette` / `createMuiTheme` know about our custom(ized) types.
  - since `grey` of `PaletteOptions` is of type `Color` from `@material-ui/core/index` (see [here](https://github.com/mui-org/material-ui/blob/master/packages/material-ui/src/styles/createPalette.d.ts)), I unfortunately had to augment that, very common, `Color` interface. The side effect is that all of [Mui's exported colors](https://github.com/mui-org/material-ui/tree/master/packages/material-ui/src/colors) now have that type. As a result, we open consumers up to a weaker typed Color value where TS doesn't throw on something like `amber.light`, but should. I think the tradeoff is acceptable for Prenda's usecases however.

I split the commits to help review.


WilliamKelley:
Closes #103

seigler (approved):

shadoath (approved):
Keeps getting more better!

PR #109, opened 2021-07-01
refactor(theme)!: separate theme properties into discrete files, rename dir to styles

WilliamKelley:
This PR is purely refactoring, my final surface level changes before getting back to real work.

- separates each property of ThemeOptions into their respective files, named by the appropriate property names
- CSSBaseline overrides now reuse from typography
- renamed '...PropOverrides' -> '...DefaultProps'
  - Mui calls them default props instead of prop overrides, i think it helps differentiate between style override by not also naming this overrides.
- renames `/theme` dir to `/styles` (mirror Mui)
- the directory only exports a single `theme` object (renamed from `muiTheme`) instead of our incomplete argument as well.
  - it's still valid for consumers to import just that `theme` object into their project, modify it, and give it to `createMuiTheme` again for their own purposes. So that justifies getting rid of `prendaTheme`.

Each commit reflects a single change.


WilliamKelley:
Apparently caused some visual changes, not my intention, resolving them now.

WilliamKelley:
Resolved changes.

seigler (approved):
Nice

shadoath (approved):

PR #110, opened 2021-07-02
feat(Pagination): initial implementation, along with PaginationItem

### LINKED ISSUE #60: Pagination
### Variations:
**Types**: `Left-Arrow`, `Center`, `More`, `Right-Arrow`
**Sizes**: NA

**Design reference:**
[Figma file - Pagination](https://www.figma.com/file/y6Ca8BdJtQEd3Of1saRieF/Spark-Design-System-Web?node-id=949%3A463)

WilliamKelley:
Includes #109 

Closes #60 

Style overrides and default props for Pagination and Pagination Item from Mui Lab.
- added `@material-ui/lab` package

The only discrepancies are that the next / previous arrow icon SVGs are internal to Mui, so not customizable. I figured it's not worth exporting a custom component to address only that. 
- I have [filed an issue](https://github.com/mui-org/material-ui/issues/27085#issue-936168933) to Mui to add that functionality though. If that's approved and added, I can add a v5 TODO for setting custom icons through default props.

The styles do get a bit weird, with negative margins and z-indexes. This is because individual `PaginationItem`'s need a 2px border. But, they ALSO need to only have a single 2px border between then when displayed in a `Pagination` group...

Adds two story groups: Pagination and Pagination Item. Both have configurable stories with sensible `argTypes` spelled out.

Adds storybook plugin package for setting pseudo states.

Pagination Item also has a story for every sensible prop combo and pseudo state. I did this because even though in our other stories, a user can change the configurable controls to see every possible variant, color, etc., BUT Chromatic snapshots won't go through every combination and tell us when one of them changes, only the default (!!) -- we have to make a story showing all of them at once, or many stories for each one. I chose to do the most of each that I could: group component /variation states and create additional stories for pseudo-versions of that group.
- because that's the only way to set the pseudo state (`Story.parameters = { pseudo: { hover: true } };`).


WilliamKelley:
The individual stories for next / previous / last / first `PaginationItem`'s don't have rounded corners because that styling has to be set in `Pagination` as first / last child -- Mui doesnt have class names for specific icon `PaginationItem`'s. I figure that's acceptable as the Pagination story will always show them and alert us to changes

seigler (approved):
Wow William! This is excellent. I hope MUI team accepts your enhancement

shadoath (approved):
Nice detailed stories

PR #112, opened 2021-07-06
feat(Checkbox)!: de-couple from FormControlLabel

WilliamKelley:
Includes #115  -- last commit contains relevant changes

Removes custom Checkbox, CheckboxProps. It was a mistake to export a custom `Checkbox` component that hard-couple the `FormControlLabel` with it. There are good reasons to keep them separate.
- Should be able to use a checkbox without a label, ex: favorite / bookmark buttons, implemented as a checkbox form group


seigler (approved):
Looks great.

shadoath (approved):

PR #113, opened 2021-07-06
feat(Radio): update styles

### LINKED ISSUE #43: Constant Radio size
It used to be that the 'Focus' Radio in the [Figma page](https://www.figma.com/file/y6Ca8BdJtQEd3Of1saRieF/?node-id=1056%3A1285) was something like 22px diameter compared to 24px in the other states. I included a [`scale(0.9)` transform](https://github.com/prenda-school/prenda-spark/blob/a8fd77ee8e5cfb2fc637d67fb7a3a9f1abb69697/libs/spark/src/form-inputs/radio.tsx#L40) as a result, but it's since been changed/fixed in Figma, so that behavior should be removed.

WilliamKelley:
Includes #111 -- last eight commits are relevant

Closes #43

Refactors style declarations to ensure more consistency, and fix unverified styles like when `checked` and `disabled` simultaneously.

Expands stories to groups of state variations, their pseudo states, with labels, and those pseudo states.


seigler (approved):
:ship:

shadoath (approved):

PR #114, opened 2021-07-06
feat(Checkbox): update styles & add indeterminate support

WilliamKelley:
Includes #112 -- last six commits are relevant

Refactors style declarations to ensure more consistency, fix unverified styles like when checked and disabled simultaneously, adds support for `indeterminate` state.

Expands stories to groups of state variations, their pseudo states, with labels, and those pseudo states.


seigler (approved):
Splendid!

shadoath (approved):
Found out you can shift click to select a range of commits in GitHub.... That's nifty.

Issue #100, opened 2021-06-29
Remove custom Input in favor of TextField


WilliamKelley:
Spark exports its own `Input` component that includes `InputLabel` and `FormHelperText` components, usually fed simply by string props. This mimics Mui's `TextField` pattern but just for `input` element purposes. Our consumers have expressed that `TextField` can be confusing to use / custom style, so that's compelling for us to keep this custom implementation.

I propose we rename `Input` to `InputField` to differentiate this from Mui's `Input` that we globally style and to give a subtle indication that this is a full-featured form field (with label and helper text).

WilliamKelley:
I'm leaning towards just removing `Input`/`InputField` in favor of `TextField`, because it so many props to destructure and pass around, it might not be worth the trouble.

WilliamKelley:
from #118 

> One kind-of major detail is that Input has no success prop, we custom implemented that. So we're just gonna have to recommend to consumers that they conditionally add className="SparkInput-success" in lieu of an actual success prop. Not pretty, but I don't think it's worth exporting a custom component yet just for that.

I'm going to experiment with TS augmenting the props interfaces, then [using a function in the style selectors](https://next.material-ui.com/styles/basics/#adapting-based-on-props) to detect the `success` prop and apply the styling. Only worry is coming from the [description of the `error` prop](https://material-ui.com/api/input-base/#inputbase-api)

> If true, the input will indicate an error. This is normally obtained via context from FormControl.

We can't add success prop handling to `FormControl` without creating a custom component for it AND exporting a custom component for each component that uses it. As I understand it, the objective of `FormControl` is browser consistency with certain style and attribute handlings, so it's critical.

This whole concern might be moot however, based on if `success` will even be used in designs. It's in our design system, but it doesn't seem to be a common design system state when looking at other systems.

WilliamKelley:
In regards to my previous comment, some initial experimentation is showing it's possible but quite quirky, ex:
- perf hit for each "style rule" / "CSS property" inline function -- the docs show it's performant, but doesn't provide metrics
- specificity is unclear -- without the inline func's, specificity is easily deduced by style declaration order. It seems that inline func's largely keep this behavior, but not for pseudo states -- bummer.
- since the prop isn't captured, it's passed down to the DOM, and emits a message that the property can only be a string. And passing a string is weird for this.

[...]

PR #118, opened 2021-07-06
feat(Input)!: remove custom implementation

WilliamKelley:
Part of #100.

Expands stories.

One kind-of major detail is that `Input` has no `success` prop, we custom implemented that. So we're just gonna have to recommend to consumers that they conditionally add `className="SparkInput-success"` in lieu of an actual `success` prop. Not pretty, but I don't think it's worth exporting a custom component yet just for that.


shadoath (approved):

seigler (approved):

WilliamKelley:
image
Chromatic snapshot tools removes margin space, but that loses capturing all of our box-shadow effect. Looking into a remedy

PR #121, opened 2021-07-07
refactor(Menu): add utility class for margin top

WilliamKelley:
Our `Menu`'s are spec'd to be 8px offset from their anchor. My previous method of adjusting the inset of the entire popover was naive -- that element is inset to the screen, not an area. A top margin is better. It has to be applied conditionally by the consuming scenario, so I defined a utility class. Ideally we could define something that covers all scenarios, but I haven't figured that yet.

### DIFF
--- a/libs/spark/src/Dropdown.tsx
+++ b/libs/spark/src/Dropdown.tsx
@@ -1,10 +1,9 @@
 import * as React from 'react';
 import {
-  Menu as MuiMenu,
+  Menu,
   MenuProps as MuiMenuProps,
   MenuItemProps as MuiMenuItemProps,
   Divider as MuiDivider,
-  Theme,
 } from '@material-ui/core';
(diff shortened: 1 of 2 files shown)


seigler (commented):
Could PaperProps be moved into MuiMenuDefaultProps to eliminate the wrapper altogether?

WilliamKelley:
> 
> 
> Could PaperProps be moved into MuiMenuDefaultProps to eliminate the wrapper altogether?

No because it's the offsetTop class should only be applied when the Menu is anchored to the bottom. But it can be anchored to any side, which would require a margin on a different side.

In a future PR when `Menu` stories are added, these possibilities should be detailed and additional utility classes will have to be made.

seigler (approved):

shadoath (approved):

PR #124, opened 2021-07-07
feat(Select): initial implementation

### LINKED ISSUE #123: Select
# ✅ Design Done
Variations: Default
Types: Default
Sizes: One
States: default, active , success, error, disabled, filled

### Design reference:
[Figma file - Select](https://www.figma.com/file/y6Ca8BdJtQEd3Of1saRieF/Spark-Design-System-Web?node-id=949%3A17)

WilliamKelley:
~Draft until required #121 and #122 are merged~

Closes #123

Implementing placeholder text and its styling is fairly tricky. It's not even really part of normal HTML. You have to give a default, disabled value option (`value=""`), an empty string as your initial state, and let Mui know you want to display that default value.

Storybook was also wigging out on controlling the `value` state by itself, so I had to add a basic hook for that.

The styling for "Placeholder" text is dynamic unfortunately, because it's not a placeholder like I said, and I couldn't write a style rule like `input[value=""] + & { }` because the `input` element comes _after_ the actually displayed `div`, and CSS doesn't have a "previous sibling" selector, or previous in general. If y'all have a suggestion that doesn't use a dynamic prop check, lmk! If not, I'll eventually submit a issue request to Mui to see if they'll re-order it. 

The `Dropdown` and `IconButton` stories changed in snapshots because I corrected the implementation of the `DownChevronIcon`,

### DIFF
--- a/libs/spark/src/Dropdown.tsx
+++ b/libs/spark/src/Dropdown.tsx
@@ -100,14 +100,6 @@ const SparkDropdownMenu = React.forwardRef<HTMLUListElement, DropdownMenuProps>(
         keepMounted
         onClose={handleClose}
         open={Boolean(anchorEl)}
-        anchorOrigin={{
-          vertical: 'bottom',
-          horizontal: 'right',
-        }}
-        transformOrigin={{
-          vertical: 'top',
-          horizontal: 'right',
-        }}
         ref={ref}
         PaperProps={{ className: 'SparkMenu-paperOffsetTop' }}
         {...other}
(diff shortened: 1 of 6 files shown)


shadoath (approved):

seigler:
That styling issue is a tough nut to crack. I think the approach you took is the best available. I hope Mui is receptive to moving the hidden input!

seigler (approved):

slingner (approved):

Issue #74, opened 2021-06-21
Improving NavBar


WilliamKelley:
Post-merge review of #64 :

Adhere to the Contributor guidelines. 
- Mui prefix naming
- Extending base component props
- Defining styling **_outside_** of the component (very important).
- Using `theme` from styling utility context, _not_ importing specific color values.

`Navbar` -> `NavBar`
`NavbarLink` -> `NavBarLink`
- reasoning: follow MUI. In a way this makes sense because Navbar is not naturally a word like Toolbar, same as AppBar, which is why AppBar is PascalCased, and so should NavBar be. Being consistent and aligned with MUI in naming will ease mental memory for consumers. Their naming scheme is already designed to ease mental memory for consumers

Hard-coding the monogram as the first child is probably not the right choice. Consumers should be able to change this depending on an app's design (ex: a future mobile app would have a hamburger menu button be the rightmost element). Specific styling of the monogram relevant to being in a `NavBar` couldbe implemented in a dedicated component, `NavBarMonogram`, or even simpler as a custom `variant` (the prop) of the `Monogram` component. Then composition is left up to the consumer.

To the same end, `NavBar` doesn't need to be anything more than a custom styled `AppBar`. It's simple enough for a consumer to import and use `Toolbar` in a standard design, and prevents a breaking change if a future design requires not using `Toolbar` as the only child.

`NavBarLink` should probably be implemented as `NavBarButton`, which can be a custom styling/variant of spark's existing `Button`.
- All elements of the NavBar will have the same, consistent look, but not all will be links (e.g. notification button).
- Buttons can be turned into links ([see docs](https://material-ui.com/components/buttons/#third-party-routing-library))
- Reasonable use case: disabling a NavBarLink. By web standards, anchor tags should never be disabled -- there's plenty of discussion out there on this, [here's a good one](https://css-tricks.com/how-to-disable-links/). But buttons can very much be disabled and communicate that to assistive tech. So, disabled styling can be advantaged using the same component you would for an enabled button link.
- Currently, the icon and text are not aligned, it should be a flex container -- Button already does this correctly.

PR #127, opened 2021-07-08
feat(NavBarLink)!: renamed to NavBarButton; improved styles

slingner:
Related to issue #74

- Updated `NavbarLink` to now be `NavBarButton`.
- Made a new story specifically for NavBarButton (took out of navbar story), added variations for all States.


libs/spark/src/NavBarItem.tsx
```diff
@@ -0,0 +1,97 @@
...
+  Button,
+  ButtonBase,
+  ButtonBaseProps as MuiButtonBaseProps,
+} from '@material-ui/core';
+import { ButtonProps } from './Button';
+import { theme } from './styles';
+// import clsx from 'clsx';
+// import { capitalize } from './utils';
+
+export type NavBarItemProps = ButtonProps;
```
WilliamKelley, line 13: `interface` not `type`, extend Mui's `ButtonProps`, omit `variant`, `color`. see CONTRIBUTING.md for a template on this

libs/spark/src/NavBarItem.tsx
```diff
@@ -0,0 +1,97 @@
...
+  },
+  label: {
+    color: theme.palette.text.onLightLowContrast,
+    fontSize: '1.125rem',
+  },
+  disabled: {},
+  focused: {},
+})(Button);
+
+// const SparkNavBarItem: FC<ButtonProps> = ({
```
WilliamKelley, line 46: Not sure if you want this to be commented on, but since this component doesn't have >1 variant, we can avoid going along the same lines as Spark's current `Button` implementation. You can use Mui's `Button` directly. Note that you're already doing this based on what you're styling above.

libs/spark/src/NavBarItem.tsx
```diff
@@ -4,13 +4,21 @@ import {
...
+  extends Omit<ButtonProps, 'variant' | 'color' | 'size'> {
+  startIcon?: React.ReactNode;
+  endIcon?: React.ReactNode;
+  children?: React.ReactNode;
+  disabled?: boolean;
+  disableElevation?: boolean;
+  disableFocusRipple?: boolean;
+  fullWidth?: boolean;
+  href?: string;
+}
```
WilliamKelley, line 21: You don't need to redeclare ones you want. You also don't need to define href, a consumer can do

```
<NavBarItem component="a" href="..." />
```
and that will fine by Mui's default implementation. Likewise for react-router
```
<NavBarItem component={RouterLink} to="..." />
```

See https://material-ui.com/components/buttons/#third-party-routing-library

slingner: Thanks, William, that's good to know about the React Router implementation. I will need to put some custom props in this interface though, correct? Maybe just children?

slingner: Without any, I'm getting an empty interface warning.

slingner: Two Issues:
1. I'm still getting an error message when trying to call `component={Link}`
2. Also trying to work out how to target the clicked state of the button. I attempted using the `aria-current='true'` in styles like Joshua had previously, but was not successful as of yet.

WilliamKelley: I wouldn't add react-router and demonstrate the integration. It's more likely to cause problems than anything. As long as we're extending Button, then we can absorb Mui's test suite and be assured that the integration will still work. I'd recommend just showing `<NavBarItem component="a" href="..." />` use case.

You won't need to add children, again that's already defined. Yeah I'm struggling how to set it equal to the other interface. Maybe ask around

libs/spark/src/NavBarItem.tsx
slingner: What's the harm of doing a type here? I'm not getting an error when using a type when using Omit

```
export type NavBarItemProps = Omit<ButtonProps, 'variant' | 'color' | 'size'>;
// export interface NavBarItemProps
//   extends Omit<ButtonProps, 'variant' | 'color' | 'size'> {}
```

libs/spark/src/NavBarItem.tsx
slingner: What's the harm of doing a type here? I'm not getting an error when using a type when using Omit

```
export type NavBarItemProps = Omit<ButtonProps, 'variant' | 'color' | 'size'>;
// export interface NavBarItemProps
//   extends Omit<ButtonProps, 'variant' | 'color' | 'size'> {}
```

WilliamKelley: I'm not sure. One consequence is that a consumer can extend a type, but they can extend an interface. I know they can union a type and all that jazz, but I feel like there must be a way to do it with interfaces. Maybe just setting the component type to be `ButtonProps`?

slingner: I haven't solved this piece but I have made a lot of progress on matching the Figma design today. Curious to hear your thoughts on its current state. I'll keep researching on extending interfaces in the meantime.

libs/spark/stories/navbar.stories.tsx
```diff
@@ -26,22 +27,22 @@ const Template = (args: NavBarProps) => (
...
     <Toolbar>
       <PrendaMonogramDark fontSize="62px" />
-      <NavbarLink href="/">
+      <NavBarItem href={`/`}>
         <HomeIconDuotone fillColor="transparent" />
         <span>Dashboard</span>
-      </NavbarLink>
-      <NavbarLink href="/" aria-current="page">
-        <CheckCircleIconDuotone />
+      </NavBarItem>
```
WilliamKelley, line 33: a

libs/spark/stories/navbar.stories.tsx
```diff
@@ -26,22 +27,22 @@ const Template = (args: NavBarProps) => (
   <NavBar {...args}>
     <Toolbar>
       <PrendaMonogramDark fontSize="62px" />
-      <NavbarLink href="/">
+      <NavBarItem href={`/`}>
         <HomeIconDuotone fillColor="transparent" />
         <span>Dashboard</span>
-      </NavbarLink>
```
WilliamKelley, line 32:
```suggestion
      <NavBarItem href="/" startIcon={<HomeIconDuotone fillColor="transparent" />}>
        Dashboard
      </NavBarItem>
```

WilliamKelley: likewise for all the rest

libs/spark/stories/navbar.stories.tsx
```diff
@@ -51,10 +52,10 @@ const Template = (args: NavBarProps) => (
...
-  <NavbarLink as="a">
+const NavBarItemTemplate = () => (
+  <NavBarItem href="/">
     <UsersIconDuotone />
     <span>Link Text</span>
-  </NavbarLink>
+  </NavBarItem>
 );
-export const BasicNavbarLink = NavbarLinkTemplate.bind({});
+export const BasicNavBarItem = NavBarItemTemplate.bind({});
```
WilliamKelley, line 61: This should be in it's own story

libs/spark/stories/navbar.stories.tsx
```diff
@@ -51,10 +52,10 @@ const Template = (args: NavBarProps) => (
 
 export const Basic = Template.bind({});
 
-const NavbarLinkTemplate = (args) => (
-  <NavbarLink as="a">
+const NavBarItemTemplate = () => (
```
WilliamKelley, line 55: Once in its own story, simplify to `Template`

libs/spark/stories/navbar.stories.tsx
WilliamKelley, line 61: Once in it's own story, simplify to `Basic`

libs/spark/test/navbar.spec.tsx
```diff
@@ -1,17 +1,17 @@
...
+import { SparkThemeProvider, NavBar, NavBarItem } from '../src';
 import { QuestionIconDuotone } from '../src/icons';
 
 describe('Button', () => {
   it('is truthy', () => {
     const { baseElement } = render(
       <SparkThemeProvider>
         <NavBar>
-          <NavbarLink href="/">
+          <NavBarItem href="/">
```
WilliamKelley, line 11: same change as in the story

WilliamKelley (changes requested):
Looking good! i'll go through the styling more carefully after this feedback is addressed.

slingner (commented):
> Looking good! i'll go through the styling more carefully after this feedback is addressed.

Updated! So much cleaner putting the `startIcon` directly in and deleting the span wrapper.

WilliamKelley (changes requested):
I thought about the name this weekend, and I think we should go with `NavBarButton`. Reason being, it's already within Mui's Button API to be an anchor tag if an href is provided. And the only other ideal that I had in mind was that an `Item` could be a parent container of whatever component made sense to be the child. But we can get what we want just from the Button.

libs/spark/stories/navbar.stories.tsx
```diff
@@ -26,22 +26,31 @@ const Template = (args: NavBarProps) => (
...
-        <RocketIconDuotone />
-        <span>Disabled</span>
-      </NavbarLink>
-      <NavbarLink style={{ marginLeft: 'auto' }}>
-        <InboxFilledIconDuotone />
-        <span>0</span>
-      </NavbarLink>
+      <NavBarItem
+        href={`#`}
+        startIcon={<HomeIconDuotone fillColor="transparent" />}
```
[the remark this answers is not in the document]
slingner: > Carry to rest of icons

Having trouble with this one, do we need to do a theme override? 

[https://stackoverflow.com/questions/62677833/change-size-of-starticon-endicon-material-ui]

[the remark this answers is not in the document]
WilliamKelley: No need for theme overrides, anything we can do there we can do here. I'll note though, I think my theme overrides to the SvgIcon might have been heavy handed and making life more difficult than necessary here in places -- but that's for another time, and not affecting this specific comment directly.

See the stylesheet below. There's a style rule with higher specificity than the rule that gets applied when `fontSize="large"` (i.e. the one setting `font-size: 2rem`). So you just have to match that winning rule to have your change take effect. We want to do that -- fixing intended prop behavior -- rather than being heavy handed in that rule by overriding every instance of an icon to be `2rem`.
image
So something like
```

  iconSizeMedium: {
    '& > *:first-child': {
      fontSize: '1.5rem', // FYI, our small/medium/large for icons is 16px/24px/32px
      '&.MuiSvgIcon-fontSizeLarge': {
        fontSize: '2rem',
      },
    },
  },
```

libs/spark/stories/navbar.stories.tsx
```diff
@@ -24,37 +25,32 @@ const UserMenu = styled.div`
...
-        <RocketIconDuotone />
-        <span>Disabled</span>
-      </NavbarLink>
-      <NavbarLink style={{ marginLeft: 'auto' }}>
-        <InboxFilledIconDuotone />
-        <span>0</span>
-      </NavbarLink>
+    <Toolbar style={{ gap: 8 }}>
+      <PrendaMonogramDark fontSize="4.5rem" />
+      <NavBarButton href="" startIcon={<HomeIconDuotone />} aria-current="page">
```
WilliamKelley, line 30:
```suggestion
      <NavBarButton href="#" startIcon={<HomeIconDuotone />} aria-current="page">
```
`href`'s shouldn't be blank. To not go anywhere, just supply a hash.

WilliamKelley (changes requested):
We can better validate styles if you expand on your stories for `NavBarButton`. 

Follow the pattern of some other components with expanded stories and create ones

- Patterns (label only, start icon plus label, label plus end icon)
- Patterns Disabled
- Patterns Hover
- Patterns Focus
- Patterns Active

Patterns is arbitrary, idk a better name to describe those three options

slingner:
> We can better validate styles if you expand on your stories for `NavBarButton`.
> 
> Follow the pattern of some other components with expanded stories and create ones
> 
> * Patterns (label only, start icon plus label, label plus end icon)
> * Patterns Disabled
> * Patterns Hover
> * Patterns Focus
> * Patterns Active
> 
> Patterns is arbitrary, idk a better name to describe those three options

I'm not fully understanding how to target the hover and focused states here? I see them being implemented in the story, but it doesn't appear to be showing those states unless you actually do the hover?  Maybe I'm missing something here.

WilliamKelley:
> I'm not fully understanding how to target the hover and focus states here? I see them being implemented in the story, but it doesn't appear to be showing those states unless you actually do the hover/focus? Maybe I'm missing something here.

You probably need to `npm i`

libs/spark/stories/navbarbutton.stories.tsx
```diff
@@ -10,9 +10,35 @@ export default {
...
+  <NavBarButton href="#" startIcon={<UsersIconDuotone />}>
+    Link Text
+  </NavBarButton>
+);
+
+export const WithEndIcon = () => (
+  <NavBarButton href="#" endIcon={<UsersIconDuotone />}>
+    Link Text
+  </NavBarButton>
+);
```
WilliamKelley, line 29: Group in a single template and make sure to spread `args` across each. then
```tsx
export const All = Template.bind({});
export const AllDisabled = Template.bind({});
AllDisabled.args = { disabled: true };
export const AllHover = Template.bind({});
AllHover.parameters = { pseudo: { hover: true } };
export const AllFoucs = Template.bind({});
AllFocus.parameters = { pseudo: { focus: true } };
```

libs/spark/stories/navbarbutton.stories.tsx
```diff
@@ -0,0 +1,88 @@
...
+    gridTemplateColumns="1fr 1fr 1fr 1fr"
+    gridGap="16px"
+    alignItems="end"
+    {...props}
+  />
+);
+
+const AllTemplate: Story = (args) => (
+  <GridContainer>
+    <span>Label</span>
```
WilliamKelley, line 60: Leave these labels off, they're not useful in the scenario

libs/spark/stories/navbarbutton.stories.tsx
```diff
@@ -0,0 +1,88 @@
...
+    <NavBarButton {...args}>Label</NavBarButton>
```
WilliamKelley, line 61: These need to be wrapped in <span>'s because the grid will stretch its children to take up their entire row/column space, but we don't want that.

WilliamKelley (changes requested):
Aside from the grid stretching issue, the stylings don't seem to match figma exactly, ex: disabled icon

slingner:
> Aside from the grid stretching issue, the stylings don't seem to match figma exactly, ex: disabled icon

Good catch on this one, just updated.

WilliamKelley (commented):
I still see styling issues with inconsistent padding, especially around the icons. They seem to have weird margins be default. For this, I'd wait until #144 is merged, then revisit. should be today.

Change your PR title from item to button.

The focus border is behaving poorly. 
    - First, since there's no border on the button by default, the appearance of a focus border causes layout shifts in surrounding content. I've fixed this in other place by setting a border on the default style, but the same color as whatever the background is supposed to be in each state/pseudo-state. Then when it needs to be a different color, like in focus, then just change `borderColor`. No layout shifts. Also requires accounting for that 2px width in things like padding.
    - Second, it seems to have a default transition from Mui on it that's just laggy. Maybe it's because of the layout shifts above and fixing that will fix this, but if not, just declare the transitions (copy paste from Mui), but without targeting border.

Lastly, it seems like if the only prescribed size for NavBarButton is 48px in height, then it should take on the default prop of `size="large"` since that's the same height. Some of your stylings will have to target different class slots after changing that, but you shouldn't have to declare a specific height as a result. Or ideally even fontsize or anything.

libs/spark/src/NavBarButton.tsx
```diff
@@ -3,16 +3,22 @@ import { withStyles } from '@material-ui/core/styles';
 import { Button, ButtonProps } from '@material-ui/core';
 import { theme } from './styles';
 
-export type NavBarButtonProps = Omit<ButtonProps, 'variant' | 'color' | 'size'>;
+export type NavBarButtonProps = Omit<ButtonProps, 'variant' | 'color'>;
```
WilliamKelley, line 6: Unless we can figure out a way to associate this with the declaration of `NavBarButton` then its useless as anything except a reference.

libs/spark/src/NavBarButton.tsx
```diff
@@ -3,16 +3,22 @@ import { withStyles } from '@material-ui/core/styles';
...
 
 export const NavBarButton = withStyles({
   root: {
-    height: 48,
     borderRadius: 8,
+    backgroundColor: 'transparent',
+    border: '2px solid transparent',
     padding: 8,
```
WilliamKelley, line 13: since border is 2px, padding has to account for that
```suggestion
    padding: 6,
```

libs/spark/src/NavBarButton.tsx
```diff
@@ -3,16 +3,22 @@ import { withStyles } from '@material-ui/core/styles';
...
     textTransform: 'none',
     color: theme.palette.text.onLightLowContrast,
-    fontSize: '1.125rem',
+    transition: theme.transitions.create(
+      ['background-color', 'box-shadow', 'color'],
+      {
+        duration: theme.transitions.duration.short,
+      }
+    ),
```
WilliamKelley, line 21: it looks like this isn't necessary after all. I tried commenting it out and it no longer lags, so it was due to the progressive layout shifts solved by a constant border size.
```suggestion
```

libs/spark/src/NavBarButton.tsx
```diff
@@ -0,0 +1,76 @@
...
+      border: `2px solid ${theme.palette.blue[4]}`,
+      outline: 'none',
+      backgroundColor: theme.palette.grey.light,
+    },
+    '&:active, &[aria-current=page]': {
+      backgroundColor: theme.palette.grey.light,
+      color: theme.palette.text.onLight,
+      '& .MuiSvgIcon-colorPrimary': {
+        color: theme.palette.text.onLight,
+        '& > path:first-child': {
```
WilliamKelley, line 43:
```suggestion
        '& > *:first-child': {
```
Some icons have `<circle>` or other svg child tags.

libs/spark/src/NavBarButton.tsx
```diff
@@ -0,0 +1,76 @@
...
+  },
+  disabled: {
+    '&.MuiButton-contained': {
+      opacity: '100%',
+      color: theme.palette.grey.dark,
+      backgroundColor: 'transparent',
+      border: 'transparent',
+      '& .MuiSvgIcon-colorPrimary': {
+        color: theme.palette.grey.dark,
+        '& > path:first-child': {
```
WilliamKelley, line 67:
```suggestion
        '& > *:first-child': {
```

libs/spark/src/NavBarButton.tsx
```diff
@@ -0,0 +1,76 @@
...
+        fontSize: '2rem',
+      },
+    },
+  },
+  disabled: {
+    '&.MuiButton-contained': {
+      opacity: '100%',
+      color: theme.palette.grey.dark,
+      backgroundColor: 'transparent',
+      border: 'transparent',
```
WilliamKelley, line 64:
```suggestion
      borderColor: 'transparent',
```
Otherwise all sub-properties of `border` are defaulted to browser defaults. You could notice the button's size reduced when switch stories from All to All Disabled.

WilliamKelley (changes requested):
See my comments.

In addition, we'll have to wait on this PR until some issues with Button are fixed. The icons are still off:
current:
image
intended:
image

But that requires another PR.

Also, we've run out of Chromatic snapshots so we can't verify everything else is staying the same, so we gotta discuss that as a team.

WilliamKelley:
Also, add another story for the `active` pseudo state

WilliamKelley:
Keeping this as Draft until we verify locally that all styles match Figma, and this PR is ready for final chromatic review

WilliamKelley:
@slingner Checking in on the status of this.

slingner:
> @slingner Checking in on the status of this.

Thanks for checking in on this William, I had to step away to work on some Create UI updates, but jumping back in now. Just pushed up changes with the active state in the NavBarButton story. 

Should I do a rebase on main also?

WilliamKelley:
@slingner Yes definitely should rebase. I'll take a look later today after that.

slingner:
Superseding by PR #166

PR #130, opened 2021-07-08
feat: new library @prenda/spark-icons

WilliamKelley:
In progress...

## 1 of 4
This is the first of four PRs that will make up an address to #73. They're separated primarily to make each sensibly reviewable, and also for respective merge commits to paint a better picture. Each PR will be marked as Draft until the previous is merged and the next can be rebased off `main`.

**1. Project generation**
2. Copying of icon utility scripts from Mui & Modification for our use case
3. Removal of icons from @prenda/spark & Specific generation and replacement of depended-on icons
4. Addition of all icons in @prenda/spark-icons

### Preface
I corrected the `"npmScope"` listed in `nx.json` from `"prenda-spark"` to `"prenda"` (no "@" needed, see @nrwl/nx's setup).

### Project generation
Generated a new NX project, namely a react component library with no styling solution.
```bash
> npx nx g @nrwl/react:lib --publishable --importPath @prenda/spark-icons
✔ What name would you like to use for the library? · spark-icons
✔ Which stylesheet format would you like to use? · none
UPDATE workspace.json
UPDATE nx.json
CREATE libs/spark-icons/.eslintrc.json
CREATE libs/spark-icons/.babelrc
CREATE libs/spark-icons/README.md
CREATE libs/spark-icons/package.json
CREATE libs/spark-icons/src/index.ts
CREATE libs/spark-icons/tsconfig.json
CREATE libs/spark-icons/tsconfig.lib.json
UPDATE tsconfig.base.json
CREATE libs/spark-icons/jest.config.js
CREATE libs/spark-icons/tsconfig.spec.json
UPDATE jest.config.js
CREATE libs/spark-icons/src/lib/spark-icons.spec.tsx
CREATE libs/spark-icons/src/lib/spark-icons.tsx
```
The exact options and steps above turned out to be necessary. All together they are what allow for projects within our NX monorepo to validly reference the icons package. Otherwise, NX will complain that we cannot import from `@prenda/spark-icons` in any file under `libs/spark/*` because "a buildable library cannot import or export from non-buildable libraries" by way of it's "enforce-module-boundaries" lint rule.
- To allow deep imports like `import ChevronDownIcon from '@prenda/spark-icons/ChevronDown`, and to specifically address the above linting errors, the last commit adds some additional rules in `tsconfig.base.json` and `.eslintrc.json`
- We primarily would use the icons in our story files, but also sometimes in component files. We'll have to see what the consequences of that are in publishing these packages -- Mui moves all icons used in the core package under `internal/` and doesn't pull in their icons package; there isn't overlap between the two for them, but there is for us, so we have a better reason.
  - We almost only use our icons in stories, which don't need to be under the spark library, and can be relocated 

### Script copying
With future knowledge, I've copied all the files from Mui that we'll need to modify, and installed only the packages we'll need later on -- the files might reference things I haven't installed, that's okay, it's meant to be in a broken state the next PR can highlight the changes clearly.

I've given permalink attribution for everything (except `templateSvgIcon.js`, that needs to be unchanged because it's literally used as a template). Mui is open-source under the MIT-license, so this is permissible. I didn't copy the download script that would make the process completed automated because our icons come from Figma and aren't hosted on the web by anyone. But, Figma does have an API that we could leverage in the future to prevent manual intervention in future maintenance (see #131 for such improvements)

### Script modifying

As outlined in `BUILDER.md`, the manual export process should end with a directory `libs/spark-icons/svg-files` with three sub directories `line/`, `fillled/`, `duotone/`.

`default/renameFilter.js` is one of the first steps in generation. The file name is `PascalCased.js` with `snake_case.svg` and `Whatever - YouCallThis 2.svg` formats being transformed. I could only find one instance of an icon name starting with a number, which is renamed to its word equivalent -- variable names can't be numbers. The sub directory of `svg-files/` that each file belongs to is also matched and added as a suffix to the filename (note, Line becomes the default).
- `svg-files/line/Activity.svg -> src/Activity.js`
- `svg-files/filled/Activity.svg -> src/ActivityFilled.js`
- `svg-files/duotone/Activity.svg -> src/ActivityDuotone.js`

Notable changes in `builder.js` are 1) stripping attributes our icons tend to come out with, that are unnecessary 2) altering the `transform` scaling logic to accept that our icon files aren't suffixed by `_24px` (even though our icons are always 24px x 24px). Instead I implemented a dynamic algorithm to scale to variable width and height values in `viewBox` (this will be useful in custom icons described in improvements here #131).

`babel.config.json` exists to be passed to the `"src:icons"` script (see `package.json`). It can be moved if it causes issues from NX's perspective, but I didn't encounter any. The config is necessary to run `builder.js` or it will produce errors about "es" modules, needing to mark "module" type and such.

Installed only required packages. Note that SVGO needed to be installed at a version <2 because of a breaking API change (improvement opportunity detailed here #131)
```bash
> npm i -D @babel/cli @babel/node @babel/plugin-proposal-class-properties @babel/plugin-transform-runtime fs-extra glob mustache path rimraf svgo@1.3.0 temp yargs
```

### Removal and Replacement
All 1700 icon and index files were removed.

Only the icons we use in components/stories were generated and their usage replaced with deep imports (which should be the standard for us so that when consumers pull in Spark with a non-tree-shaking bundler, they face no difficulties -- or we can transform imports during our build process -- might be a good thing to pursue regardless).


WilliamKelley:
Closed in favor of #132 and onwards

PR #132, opened 2021-07-09
chore(nx): generate new project, spark-icons (@prenda/spark-icons)

WilliamKelley:
## 1 of 4
This is the first of four PRs that will make up an address to #73. They're separated primarily to make each sensibly reviewable, and also for respective merge commits to paint a better picture. Each PR will be marked as Draft until the previous is merged and the next can be rebased off `main`. You can effectively review each PR by commit.

**1. Project generation (this PR)**
2. Copying of icon utility scripts from Mui & Modification for our use case (#133)
3. Removal of icons from @prenda/spark & Specific generation and replacement of depended-on icons (#134)
4. Addition of all icons in @prenda/spark-icons (#135)

### Preface
I corrected the `"npmScope"` listed in `nx.json` from `"prenda-spark"` to `"prenda"` (no "@" needed, see @nrwl/nx's setup).

### Project generation
Generated a new NX project, namely a react component library with no styling solution.
```bash
> npx nx g @nrwl/react:lib --publishable --importPath @prenda/spark-icons
✔ What name would you like to use for the library? · spark-icons
✔ Which stylesheet format would you like to use? · none
UPDATE workspace.json
UPDATE nx.json
CREATE libs/spark-icons/.eslintrc.json
CREATE libs/spark-icons/.babelrc
CREATE libs/spark-icons/README.md
CREATE libs/spark-icons/package.json
CREATE libs/spark-icons/src/index.ts
CREATE libs/spark-icons/tsconfig.json
CREATE libs/spark-icons/tsconfig.lib.json
UPDATE tsconfig.base.json
CREATE libs/spark-icons/jest.config.js
CREATE libs/spark-icons/tsconfig.spec.json
UPDATE jest.config.js
CREATE libs/spark-icons/src/lib/spark-icons.spec.tsx
CREATE libs/spark-icons/src/lib/spark-icons.tsx
```
The exact options and steps above turned out to be necessary. All together they are what allow for projects within our NX monorepo to validly reference the icons package. Otherwise, NX will complain that we cannot import from `@prenda/spark-icons` in any file under `libs/spark/*` because "a buildable library cannot import or export from non-buildable libraries" by way of it's "enforce-module-boundaries" lint rule.
- To allow deep imports like `import ChevronDownIcon from '@prenda/spark-icons/ChevronDown`, and to specifically address the above linting errors, the last commit adds some additional rules in `tsconfig.base.json` and `.eslintrc.json`
- We primarily would use the icons in our story files, but also sometimes in component files. We'll have to see what the consequences of that are in publishing these packages -- Mui moves all icons used in the core package under `internal/` and doesn't pull in their icons package; there isn't overlap between the two for them, but there is for us, so we have a better reason.
  - We almost only use our icons in stories, which don't need to be under the spark library, and can be relocated

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -10,7 +10,7 @@
           "error",
           {
             "enforceBuildableLibDependency": true,
-            "allow": [],
+            "allow": ["@prenda/spark-icons", "@prenda/spark-icons/*"],
             "depConstraints": [
               {
                 "sourceTag": "*",
--- a/jest.config.js
+++ b/jest.config.js
@@ -1,3 +1,3 @@
 module.exports = {
-  projects: ['<rootDir>/libs/spark'],
+  projects: ['<rootDir>/libs/spark', '<rootDir>/libs/spark-icons'],
 };
--- a/libs/spark-icons/.babelrc
+++ b/libs/spark-icons/.babelrc
@@ -1,0 +1,12 @@
+{
+  "presets": [
+    [
+      "@nrwl/react/babel",
+      {
+        "runtime": "automatic",
+        "useBuiltIns": "usage"
+      }
+    ]
+  ],
+  "plugins": []
+}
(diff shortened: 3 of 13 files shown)


seigler (approved):
Thank you for getting to the heart of the issue! This is very valuable.

shadoath (approved):

PR #135, opened 2021-07-09
feat(spark-icons): add all icons

WilliamKelley:
## 4 of 4

Closes #73. Reviewable by commit.

1. Project generation (#132)
2. Copying of icon utility scripts from Mui & Modification for our use case (#133)
3. Removal of icons from @prenda/spark & Specific generation and replacement of depended-on icons (#134)
4. **Addition of all icons in @prenda/spark-icons (this PR)**

### Addition
All 1688 icon files were generated and added, in addition to the 9 existing files

### Building
The library is buildable, which is great. There's an error about `templateSvgIcon.tsx` not being under `src/`. The build generates a typing file for it. I would like to have it just be excluded from the build process, but I don't know how to specify that.
```bash
> nx run spark-icons:build 
Bundling spark-icons...
rpt2: options error TS6059: File '/home/william/projects/spark/libs/spark-icons/defaults/templateSvgIcon.tsx' is not under 'rootDir' '/home/william/projects/spark/libs/spark-icons/src'. 'rootDir' is expected to contain all source files.
'elementType' is not exported by 'node_modules/prop-types/index.js'
No name was provided for external module 'react-is' in output.globals – guessing 'reactIs'
┌────────────────────────────────────────────────────────────────────────────────────────┐
│                                                                                        │
│   Destination: /home/william/projects/spark/dist/libs/spark-icons/spark-icons.umd.js   │
│   Bundle Size:  2.51 MB                                                                │
│   Minified Size:  2.21 MB                                                              │
│   Gzipped Size:  391.99 KB                                                             │
│                                                                                        │
└────────────────────────────────────────────────────────────────────────────────────────┘
Bundle complete: spark-icons
rpt2: options error TS6059: File '/home/william/projects/spark/libs/spark-icons/defaults/templateSvgIcon.tsx' is not under 'rootDir' '/home/william/projects/spark/libs/spark-icons/src'. 'rootDir' is expected to contain all source files.
'elementType' is not exported by 'node_modules/prop-types/index.js'
┌────────────────────────────────────────────────────────────────────────────────────────┐
│                                                                                        │
│   Destination: /home/william/projects/spark/dist/libs/spark-icons/spark-icons.esm.js   │
│   Bundle Size:  2.47 MB                                                                │
│   Minified Size:  2.28 MB                                                              │
│   Gzipped Size:  395.78 KB                                                             │
│                                                                                        │
└────────────────────────────────────────────────────────────────────────────────────────┘
Bundle complete: spark-icons

[...]

PR #139, opened 2021-07-12
feat(spark-icons): build to be path-import friendly without tree-shaking

### LINKED ISSUE #138: Making @prenda/spark-icons tree-shakable isn't enough
See [option 1 of Mui's section on bundle size](https://material-ui.com/guides/minimizing-bundle-size/#option-1). Notice that path imports avoids bundling unused modules, without any mention of the package being tree-shakable. This is because a build of Mui _does not_ assemble every component/file into a single output file that's marked tree-shakable. Instead, individual files are maintained.

During build testing of `@prenda/spark-icons` in #135, the build output is able to be marked as tree-shakable, but individual files are not maintained. In bundlers without tree-shaking, like Meteor, this kind of a build output would still cause the same performance issues as a non-tree-shakable bundle.

To alleviate this, we will like have to build `@prenda/spark-icons` using commands and config that mirror what Mui does, preserving individual files.

WilliamKelley:
## 5 of 4

Followup to #135
Closes #138 

Using Nx, `nx run spark-icons:build` produces a tree-shakable bundle, but does not produce a bundle that's ideal for path imports and use with non-tree-shaking bundlers, like Meteor.

The resolution is to again copy and modify what Mui does.

The result is that `@prenda/spark-icons` should now be built from the command `npm run build:icons` issued at the repo root. The build output is sent to `dist/libs/spark-icons-alt` since the Nx build currently occupies the `dist/libs/spark-icons` path.

### Visual representation
`> nx run spark-icons:build` produces _n_ TS declaration files BUT _1_ file for each bundle type
image
image

vs.

`> npm run icons:build` produces _n_ TS declaration files AND _n_ files
image
image

### DIFF
--- a/libs/spark-icons/defaults/babel.config.js
+++ b/libs/spark-icons/defaults/babel.config.js
@@ -19,18 +19,48 @@ if (process.env.BABEL_ENV === 'es') {
   ];
 }
 
+const productionPlugins = [
+  '@babel/plugin-transform-react-constant-elements',
+  'babel-plugin-transform-dev-warning',
+];
+
 module.exports = {
-  presets: defaultPresets,
-  plugins: [['@babel/plugin-proposal-class-properties', { loose: true }]],
+  presets: defaultPresets.concat([
+    '@babel/preset-react',
+    '@babel/preset-typescript',
+  ]),
+  plugins: [
+    ['@babel/plugin-proposal-class-properties', { loose: true }],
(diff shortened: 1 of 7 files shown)


shadoath (approved):
5/4 More that 100% awesome here!

seigler (approved):
Excellent

WilliamKelley:
Merging without Chromatic UI Review / Tests because no Spark `/src` files are affected by this

Issue #101, opened 2021-06-29
Button


WilliamKelley:
Noticed these changes in the [Button Figma page](https://www.figma.com/file/y6Ca8BdJtQEd3Of1saRieF/Spark-Design-System-Web?node-id=29%3A0)

Variants (appearance): `solid`, `outlined`, `flat` -> `primary`, `secondary`, `ghost`

Default size: `medium` -> `large`

New 'Active' spec

New 'Loading' spec

@ssbittle Any other changes you know of?

ssbittle:
@WilliamKelley Active state
<img width="134" alt="CleanShot 2021-06-30 at 08 30 44@2x" src="https://user-images.githubusercontent.com/83429562/123969074-790d8280-d97d-11eb-8a04-c9c4a05bc720.png">

WilliamKelley:
@ssbittle I'd like to propose a change to the `variant` names. I know they've already undergone a change, but I think `primary` and `secondary` are less intuitive values for these stylings, at least in a dev's eyes.

Mui's default variants are `contained`, `outlined`, and `text`. They reserve `primary` and `secondary` to describe the _color_ options that can be combined with any variant. This stems from their palette having clear primary and secondary values, which Spark's palette does not however.

Still, I think it would be valuable for Spark to adopt those variant values. Outlined is the mental image I have of that Button styling, whereas secondary doesn't inspire any mental intuition. Likewise for the others.

Adopting Mui's values would also simplify our implementation greatly, as we customize the existing interface and avoid exporting a custom interface.

ssbittle:
@WilliamKelley I'm on board!

WilliamKelley:
Changes have been updated in #158 except for loading spec, but that can be tracked in an isolated issue #164 .

PR #142, opened 2021-07-12
feat(Button): update stylings

WilliamKelley:
Part of #101 

- _Does not update the variant names to the new values_
- Adds `:active` styling
- Updates default values.
- Expands the stories to showcase all combinations of the Button's props.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,0 +1,1 @@
+/libs/spark/src/icons/*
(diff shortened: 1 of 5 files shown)


seigler (approved):
You're like a machine with these PRs!

shadoath (approved):

PR #145, opened 2021-07-12
feat(IconButton)!: remove custom implementation in favor of overrides

WilliamKelley:
Draft pending variant discussion outcome in #101

- Removes custom IconButton implementation
- Adds theme-level style overrides and default props for IconButton
- Updates internal usages of IconButton to be from Mui.

The disadvantage is that Mui's `IconButton` API does not account for use of `variant`'s like it does for `Button`. So, utility class names are leveraged instead: `"variantContained"`, `"variantOutlined"`, `"variantText"`. This is not ideal, so please give feedback if you think it's acceptable or not.


WilliamKelley:
Keeping this a draft because I think we should pursue exporting a custom wrapper to extend the props for this. Asking consumers to memorize two class utils isn't a good precedent imo.

WilliamKelley:
superceded by #173

PR #141, opened 2021-07-12
feat(spark-icons): add illustrations and logos

WilliamKelley:
## 6 of 4

Followup to #135 and #138.
Addresses part of #131.

Reviewable by commit.

Adds Illustration and Logo svgs to `@prenda/spark-icons`. Restores a pattern from Mui of maintaining custom / legacy icons that are copied into each generation of the larger set of icons.

Allows for Illustrations and Logos to be removed from `@prenda/spark` in #146

PR #146, opened 2021-07-12
feat(spark)!: remove illustrations and logos

WilliamKelley:
Followup to #141 

Only last commit is relevant.

- Removes exports of illustrations and logos.
- Replaces usage with imports from `@prenda/spark-icons`.
- Improves stories look.
  - As you can see from the inline styles I had to set, the usage is still less than ideal for consumers. I think improving / adding utilities to `SvgIcon` in the future could greatly ease this use-case.

This will be blocked until we can get Chromatic review.


seigler (approved):
> improving / adding utilities to SvgIcon in the future

I agree

shadoath (approved):

PR #168, opened 2021-07-22
chore(gha): add steps to publish @prenda/spark-icons

WilliamKelley:
We should be able to use the same "automation" token that's currently setup.

Builds using the custom babel config.

Notifies the slack channel separately from `@prenda/spark` publish. Maybe another reason to create a "webhooks" channel.

### DIFF
--- a/.github/workflows/npm-publish.yml
+++ b/.github/workflows/npm-publish.yml
@@ -43,3 +43,32 @@ jobs:
           MSG_MINIMAL: true
           SLACK_TITLE: '@prenda/spark'
           SLACK_MESSAGE: Failure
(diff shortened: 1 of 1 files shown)


seigler (approved):

shadoath (approved):

shadoath:
I like the idea of a webhooks channel.

WilliamKelley:
> 
> 
> I like the idea of a webhooks channel.

Great, I've discussed it in the past as well and got a bit of approval then too, no disapproval. So, I went ahead and created #spark-webhooks and a updated the slack integration webhook

PR #171, opened 2021-07-28
fix: overlapping class name generation with consumers

### LINKED ISSUE #170: `withStyles` generates overlapping class names between Spark and consumers
@seigler and @slingner found this issue.

The `withStyles` utility generates dynamic class names in production that follow a deterministic pattern: `jss1`, `jss2`, etc. following order of evaluation. Spark and PW were following different instances of the ordering, so `jss1` was declared twice and both style rules were being applied to each element.

The solution was to advantage`createGenerateClassName`, explainer [here](https://material-ui.com/styles/advanced/#class-names), api doc [here](https://material-ui.com/styles/api/#creategenerateclassname-options-class-name-generator). Confirms a past guess of best behavior for Spark in a [past issue comment](https://github.com/prenda-school/prenda-spark/issues/67#issuecomment-860734331). 

Note: this might be solved by an offshoot of #169, but we should absorb this practice nonetheless.

WilliamKelley:
Closes #170.

### DIFF
--- a/libs/spark/src/SparkThemeProvider.tsx
+++ b/libs/spark/src/SparkThemeProvider.tsx
@@ -9,7 +9,6 @@ import { theme } from './styles';
 
 const generateClassName = createGenerateClassName({
   productionPrefix: 'spark',
-  seed: 'spark',
 });
 
 export const SparkThemeProvider = (props) => {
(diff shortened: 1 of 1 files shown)


WilliamKelley:
Hm, I might have been wrong / overstepped in setting the seeds. all over our MuiComponent classes are prefixed with `spark-` now too. I'll see if not setting it avoids that, we're really just looking to make the `jssX` ones unique, otherwise all our more custom overides like '& MuiSvgIcon-root' will have to be changed to '& spark-MuiSvgIcon-root'.

seigler (approved):

shadoath (approved):

PR #173, opened 2021-07-28
feat(IconButton)!: update prop values, favor overrides, update styles

### LINKED ISSUE #172: IconButton rounded to circular
Noticed icon buttons have 50% border radius.

Also, it wasn't separated from the Button change issue in the past, but the variant names have changed as well

[Figma page](https://www.figma.com/file/y6Ca8BdJtQEd3Of1saRieF/Spark-Web-components?node-id=1690%3A59)

WilliamKelley:
Closes #172 .

- Updated old variants: "primary", "secondary", "ghost" -> "contained", "outlined", "text" (mirrors Button)
- Updated border radius styling
- Removes re-implementation of IconButton with ButtonBase. Instead supplies theme style overrides & default props. Still exports custom wrapper since the `size` prop needs different values (only `small` and `medium` by default). And the `variant` prop needed to be added entirely.


WilliamKelley:
Some IconButton inheritors (like Checkbox, Radio) are also getting the theme overrides, working on a fix now.

shadoath (approved):

seigler (approved):

PR #182, opened 2021-08-09
feat(spark typography): match Figma font weights

### LINKED ISSUE #33: Typography weights
Re-validate font-weight's since Figma was previously inconsistent

_[Figma page](https://www.figma.com/file/w9lodWeivyS4GHdBNIYaJr/Spark-Design-System-Base?node-id=1%3A3)_

_[Migrated from Asana](https://app.asana.com/0/1200012653024704/1200260683928154/f)_

seigler:
Update Typography font-weights and storybook to match Figma specifications [here](https://www.figma.com/file/KTeHk612CXRDK2nrOI1nei/Spark-Typography?node-id=2%3A2)

Resolves #33

### DIFF
--- a/libs/spark/README.md
+++ b/libs/spark/README.md
@@ -35,6 +35,8 @@ Now you can develop components within the `libs/spark/src/` directory and see ch
 
 Depending on your preferred IDE, consider downloading a [community IDE plugin for Nx workspaces](https://nx.dev/latest/react/getting-started/console#download).
 
+The Spark theme refers to font files for _Nunito_, which should be served at the `/fonts` path. These fonts are available in this library in `libs/spark/public/fonts`.
+
 ## Testing
 
 Run tests with `nx run spark:test`.
--- a/libs/spark/src/Typography.tsx
+++ b/libs/spark/src/Typography.tsx
@@ -14,7 +14,7 @@ const styles = (theme: Theme) =>
   createStyles({
     root: {
       '& strong, b': {
-        fontWeight: 500,
+        fontWeight: 600,
       },
     },
     initial: {}, // to satisfy TS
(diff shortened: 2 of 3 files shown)


WilliamKelley:
I think you'll need to set the extrabold font-face at-rules in `src/styles/typography`. I think we should move to specifying _all_ font-weights and supply all files regardless of what our typography ranges.

shadoath (approved):

WilliamKelley (changes requested):
see previous comment

seigler:
@WilliamKelley it looks like extraBold is already present in the AtRules in `typography.ts`. I do see a related typography problem which might be reasonable to include in this PR: in Prenda World, the Spark css is trying to load Nunito from _Prenda World's_ web server filesystem, e.g. `localhost:3000/fonts/nunito-medium-webfont.woff2` in a dev environment. Is there some file-path utility we should be using in `typography.ts` to create those paths?

WilliamKelley:
Are they not present? How is PW loading font files?

Those rules make the assumption the consumer will provide the files at the /fonts/... static path of their application. Which is usually the standard and so is a fine assumption. I thought PW was going to do that.

But if not, we can open that to customization.

seigler:
Prenda World currently just has a Google Fonts `link` tag to https://fonts.googleapis.com/css2?family=Nunito:ital,wght@0,200;0,300;0,400;0,600;0,700;0,800;0,900;1,200;1,300;1,400;1,600;1,700;1,800;1,900&display=swap
and doesn't have a local copy of the font files. The Spark definition overwrites the Google definition, but since the Spark definition's path is missing locally, no Nunito is available. It's a beautiful little mess :joy:

WilliamKelley:
Gotcha. Looks like that will load 200 and 300 weights as well, but i dont think any designer goes that light so we don't have to add rules for them and everything below 400 will default to 400.

For a meteor app, PW can just copy the files that Spark has into its /public folder and meteor will bundle and serve those normally. As is the case with practically every bundler/server.

I meant to add that in the documentation way back when, but that was before we scrapped a separate docs site. Definitely could add to the README or to a Story in the Typography Storybook group since that's kind of our living docs ?

Additionally, if PW doesnt want to bundle fonts itself, there can be two solutions: 
  - we build a dedicated opt-out feature for those font rules in SparkThemeProvider
  - or, the consumer simply doesnt use SparkThemeProvider, and instead re-creates or modifies the theme to remove those fontface atrules and re-builds their own provider.

I'm fine with any options as long as it's not unnecessarily complex for Spark.

seigler:
I merged a change to Prenda World that makes the font files available there, and added a README line here in `spark` that mentions the expectation for Nunito fonts at `/fonts`.

WilliamKelley (approved):

PR #184, opened 2021-08-09
fix(spark-icons): camelCase hyphenated path attrs

### LINKED ISSUE #183: Some spark icons have invalid `path` attributes
Some path attributes in `spark-icons` are hyphenated instead of camelCased, producing React errors.

BeakerDuotone is one example:
> Warning: Invalid DOM property `stroke-linecap`. Did you mean `strokeLinecap`?

seigler:
Resolves #183


WilliamKelley:
Although I'd like the fix to be in Figma, it'll be fine to monkey patch since its only our few custom icons.

But you'll also need to modify the icons builder script to camel case these attributes if it finds them on the svg. I think it's clear in the file where that's done. That way in the future, we won't have to monkey patch them again.

WilliamKelley (changes requested):
see previous comment

WilliamKelley (approved):

shadoath (approved):

PR #188, opened 2021-08-13
feat(theme): adjust shadows to new elevations spec

WilliamKelley:
Translates Spark Design Elevations E100 to E500 to `theme.shadows[1 ... 5]` and sets `theme.shadows[5 ... 24]` to be equivalent to `theme.shadows[5]`, the highest elevation shadow.

Replaces Mui's utility with one that fits our exact need.

Adjusts components that use elevations (Card and Menu) to their new prescribed elevations.

Unfortunately, it was not possible to customize the shadows object to have our elevations correspond to keys like `shadows[100]` etc. This was because the `type Shadows` cannot be overridden with TS module augmentation. Other, more cruder overriding methods still couldn't satisfy dependents, like `ThemeOptions` or `Theme`, to play well with a custom type.

### DIFF
--- a/libs/spark/src/Card.tsx
+++ b/libs/spark/src/Card.tsx
@@ -1,8 +1,9 @@
 export { Card } from '@material-ui/core';
 
 export const MuiCardDefaultProps = {
-  elevation: 4,
+  elevation: 2, // E200
 };
+
 export const MuiCardStyleOverrides = {
   root: {
     borderRadius: 16,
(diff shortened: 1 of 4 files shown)


WilliamKelley:
Hm, the shadows seem to have disappeared from Card and Menu, debugging now

WilliamKelley:
Resolved, ready for review again

slingner (approved):

seigler (approved):
This is still a totally usable solution. Nice.

PR #194, opened 2021-08-13
refactor: re-generate font-files, revise font face at rules

### LINKED ISSUE #193: Reconcile font files and font-face at-rules
First issue: there's a "Medium" font-face at-rule and font-file for Nunito. However, there's no Medium spec on [Google's font site](https://fonts.google.com/specimen/Nunito) or in [their source repo](https://github.com/googlefonts/nunito/tree/main/fonts/TTF). I believe it came from the copy that [font-squirrel makes available](https://www.fontsquirrel.com/fonts/nunito). Either way, I think the presence of a Medium variant is wrong. It's also not used by any of our Typography's, and from font-squirrel's example, does not differ in design from Regular.

Second issue: The typography theme file, `src/styles/typography.ts` has at-rules for Semibold, but links the "bold" fonts. This is a mistake.

Third issue: the Source Code Pro files seem to have disappeared. Not sure at what point in history this happened.

WilliamKelley:
Closes #193 

Things this PR does
- Remove Nunito Medium. As explained in the issue, it's redundant.
- Re-generated Nunito files, downloaded from Google, converted to `.woff(2)` through FontSquirrel.
- Restore Source Code Pro fonts
- Supply and use a custom `pxToRem` utility function through theme typography.
  - mui's default pxToRem takes the default size as 14, but we want 16. 
- correct the `Semibold` font face at rule objects to load the `semibold` files instead of bold.
  - i think i originally did that because Spark didn't prescribe a 600 semibold? I'm not sure, the early days of Spark fonts were too volatile and things are much more standard now. These changes are the way things should be.
- Adds to `vNext` changelog

Additionally, I've revised some component stylings that were supposed to have 600 font weight, but had 500 because previously 600 was too heavy (since it was really 700).
  - In chromatic you'll see _very slight_ changesets for these, FormControlLabel, InputLabel, use of Typography

And lastly, I removed the custom `tertiaryBlue2` Typography color that was used in the typography story. I don't think we ever should have outright supported it, I replaced it with a one-off styling in the story, which is what it should have been. Along those lines, I don't consider it's removal a breaking change.

It's unfortunate this PR grew so large in the changes, I apologize. If i had split things up, we would have had to accept significant, bad changesets in Chromatic for each PR. I figured this was better.


seigler (approved):
Fantastic.

shadoath (approved):
You really went at it.

Does this also help with lining up our Figma text sizes to match what we enter into the code?
Good to have the pxToRem method.

WilliamKelley:
> Does this also help with lining up our Figma text sizes to match what we enter into the code?

Do you mean an easy way to use typography specs in style objects? => the next PR (#195) will address that

Or translating pixels to rem? => `pxToRem` does that yes

PR #195, opened 2021-08-14
feat(theme): add custom typography keys

### LINKED ISSUE #191: Advantage typography theme property more
There are many places where it would be useful easily apply one of our typography values. For example, the upcoming Avatar uses a "Heading/Large" typography instance for textual children, but we would want the styles to just be set on that containing `div` tag, and not instantiate `Typography`. Hence in a perfect world we would want
```jsx
const CustomAvatar = withStyles((theme) => ({
  root: {
    ...theme.typography['heading-lg']
  }
}))(Avatar);
```
and have the `font-size`, `font-weight`, `line-height` spread.

In fact, Mui already enables this behavior with their built-in variants

```jsx
const CustomAvatar = withStyles((theme) => ({
  root: {
    ...theme.typography.h4 // or .h1 thru .h6, .body, .body2, etc.
  }
}))(Avatar);
```

We should look into if we can add our custom keys straight to the typography theme parameter (`src/styles/typography.ts`). If not, we'll just have to match our existing `variantMapping` determined in `Typography.tsx`.

WilliamKelley:
Depends on #194 

Closes #191 

`theme.typography` has the ability to provide theme consumers access to theme-level typography customizations. This PR not only carries over the Spark-to-Mui variant equivalents previously only defined in Typography.jsx (e.g. `h1, h2, ... h6, body1, body2, etc.`), but also augments the necessary `createTypography` module interfaces so that we can spread our custom variants (e.g. `heading-lg, heading-md, etc.`) over `theme.typography` and have consumers be able to access it through any use of theme, all while keeping TS happy :D.

The "strong" variants of labels were added as well, because they are so frequently used in other components' styling. The one part i want to emphasize is that a "Strong Label" is _not_ a replacement for actually using a `<strong>` tag around text that needs to be emphasized. The `<strong>` tag has semantic importance that is highlighted to assistive technology as emphasized potion of larger text. On the other hand, the `font-weight` CSS property is _purely stylistic_.

Similar to what I did in shadows, I customized a `buildVariant(...)` utility that I think helps keep the declarations very clean and focused.

Subsequent usages of typography values in components were also adjusted -- primarily Typography, but lots of components use a `label-...` variant.
  - There's a non-trivial amount of changes to several components, like `Button`, `NavBarButton`, `InputBase/Select/TextField` because this highlighted inconsistencies among our font definitions that brought out layout changes that this PR now also had to fix.


WilliamKelley:
Blocked by #202 currently

shadoath (approved):

libs/spark/src/NavBarButton.tsx
```diff
@@ -30,11 +30,27 @@ export const NavBarButton = withStyles((theme) => ({
...
+    marginTop: '-2px',
+    marginBottom: '-2px',
+    marginLeft: '8px',
+    marginRight: 0,
   },
   iconSizeLarge: {
-    // override Mui selector
-    '& > *:first-child': {
+    // artificially increase specificity to win battle with default
+    '&& > :first-child': {
```
seigler, line 53: That's a nice trick!

seigler (approved):

libs/spark/src/NavBarButton.tsx
WilliamKelley: right! but a dirty trick.

PR #196, opened 2021-08-14
feat(Typography)!: simplify color values to new palette

WilliamKelley:
Depends on #195 

Simplifies color values like `textOnLight` to `onLight`, etc.

Also fixes the "swallowing" of `classes`. Essentially, if a consumer were to customize a "class" (from the component's Mui CSS API, using a utility like `withStyles`) that our `styles` didn't declare in in `Typography.tsx`, then our implementation would simply not apply that `classes` attribute / class name. To prevent that swallowing, we have to re-define some stuff. Always sucks to copy, but this is probably an ideal customized component implementation because now consumers can override our classes like `paragraph-md` as if they were part of the CSS API. So we should think about extending this pattern to our other custom components.


WilliamKelley:
Closed in favor of #197

PR #200, opened 2021-08-16
fix(Typography): swallowing classes & missing overridable component

### LINKED ISSUE #198: [Typography] swallows non-custom/root `classes`
#### Problem
If a consumer were to customize a "class" (from the component's Mui CSS API, using a utility like `withStyles`) that our styles didn't declare in in `src/Typography.tsx`, then our implementation would simply not apply that classes attribute / class name. 

#### Solution
Originally I thought we'd have to re-define all the classes we want, which means copying, blegh. BUT, i realized that Mui's component only applies `classes` keys that it knows as part of it's CSS API. So we can just pass `classes` through to the base component even if it contains our custom keys, and even with us applying our custom keys like it is currently.

This is probably an ideal customized component implementation because now consumers can override our classes, like `paragraph-md`, as if they were part of the CSS API. So we should think about extending this pattern to our other custom components.

### LINKED ISSUE #199: [Typography] Can't use `component` prop.
```tsx
<Typography variant="heading-lg" component="h2" />
```

The above produces a TS error saying component doesn't exist on the interface.

I consider this functionality of Typography to be _essential_ to its usefulness. Styling is separate from semantics and Mui's API encapsulates that ideal perfectly.

Now, fixing this will be non-trivial as it comes down to using the `OverridableComponent` generic type from Mui. [Here's the respective file](https://github.com/mui-org/material-ui/blob/master/packages/material-ui/src/OverridableComponent.d.ts), and [here's its use in Typography](https://github.com/mui-org/material-ui/blob/master/packages/material-ui/src/Typography/Typography.d.ts). I've stared at these series of types and interfaces for a while and played around with them and I still don't understand them completely. However, I'm comfortable in saying that we're not gonna find a simpler way than they've got. So fixing this bug will involve significant copying / repurposing of their methodology -- nothing new for us though.

Solving this problem will be essential for our custom implementations as overridable components are a fantastic feature of a UI library that we should not break.

WilliamKelley:
Two major fixes that are slightly tied together

#### Swallowing `classes`
Several considerations factor into this implementation.
  - There are two categories of `classes` keys (they should be _mutually exclusive_)
    - base/underlying keys defined by Mui (i.e. the CSS API)
    - custom keys defined by Spark
  - There are two general sources of `classes`:
    - from props
    - from Spark's base customization
  - `classes` from props can contain customizations for both underlying Mui class keys and custom Spark class keys.

In the body of our custom component, we must extract the custom class keys from `props.classes` and concatenate them to our base, custom `classes`. All other keys are consequently not-custom, and transferred to an "underlying" `classes` object meant exactly for the component being extended. We do all the work because the underlying component will spit out warnings in development when it is passed `classes` with keys that aren't in its CSS API.

Once we do the extraction combination work, we're faced with an additional consideration of where do we apply our custom classes? For a component as simple as Typography, we could stick with what we've done with combining it in a class name. But, that does _not_ work when the `classes` structure of the underlying component doesn't apply classes only to the parent component. For instance,
```jsx
function MuiComponent({ classes, className }) {
  return (
    <div className={clsx(className, classes.root, { [classes.outer]: conditional })}>
      Something with one style.
      <div className={clsx({ [classes.inner]: conditional })}>
        Something with another style.
      </div>
    </div>
  );
}
```

In that case, we have to map custom classes to the underlying classes that they are extensions of. Hence, you see the motivation behind 
```jsx
        classes={{
          ...underlyingClasses,
          root: clsx(underlyingClasses.root, {
            [customClasses[variant]]: variant !== 'inherit',
            [customClasses[`color${capitalize(color)}`]]: color !== 'initial',
          }),
        }}
```

Also, providing the custom name of `'SparkTypography'` to the `options.name` argument of `makeStyles` has an interesting further application that we can decide to pursure. After looking into the logic of Mui's styling utilities, I found that the `theme.overrides` styles are checked to see if they contain the `options.name`. If so, the classes are generated without identifiers, e.g. `SparkTypography-label-xl` and not `SparkTypography-level-xl-12`. Now there is no such key present in overrides, but I point it out to say that we could extend the overrides TS interface and add them for our components in order to do so and have class names generate more cleanly. It's not pressing right now though.

Closes #198 

#### Overridable Component interface
  - I've added a lot of TS stuff that looks very foreign to the project thus far. Indeed it is, but it's a solution to the difficult problem of allowing a consumer to specify what component/tag should be rendered -- difficult because the valid props can change based on this (an anchor tag should allow/suggest `href`, but a span tag should not).
  - `TypographyTypeMap` is very close to its Mui counterpart (see issue description link below) except for our custom props / omission of standard props
  - `TypographyProps` is identify to Mui
  - `TypographyClassKey` is extended from Mui with our custom CSS API.
  - The component is typed identical to Mui, `OverridableComponent<TypographyTypeMap>`
  - Part of the motivation for calling `make/useStyles` is from above, but is also because `withStyles` was ruining the interface of the component, saying it's not overridable.
  - In the end, `component` works exactly and intelligently as expected, and all valid `classes` are correctly identified by TS (typing `<Typography classes= {{ ''` will bring up all the valid classes api we're defining, custom + underlying... beautiful). 

Closes #199 

Also, in fixing the latter issue and bringing our custom implementation so close to Mui's, I've included ref forwarding -- something we should probably do for all components (that's three things now!).


WilliamKelley:
Forgot to add changelog entry, doing that now

libs/spark/CHANGELOG.md
```diff
@@ -13,25 +13,28 @@
   - Corrected height.
 - **NavBarButton**
   - Corrected height when no start/end icon given.
+- **Typography**
+  - TS interface accepts the `component` prop.
```
seigler, line 17: Will Ferrell as Buddy the Elf bursting into a coffee shop, subtitled "You did it!"

seigler (approved):
This Typescript stuff is starting to feel a little more magical than I'm used to. I like it.

shadoath (approved):

PR #201, opened 2021-08-16
feat(Avatar): initial implementation

### LINKED ISSUE #52: Avatars
### Variations:
**Types**: `Image`, `Icon`, `Image`
**Sizes**: Scalable

**Design reference:**
[Figma file - Avatar](https://www.figma.com/file/y6Ca8BdJtQEd3Of1saRieF/Spark-Design-System-Web?node-id=949%3A465)

WilliamKelley:
Closes #52 

- Implements a custom classes CSS API like `Typography`, that is style-able with customers using the standard utilities (`withStyles`, etc.).
  - See #200 for details
- Implements 4 sizes across 3 content types
- Content types are intentionally not variants, they have intertwining behavior, see Mui docs.
- Adds documentation story linking out to Mui and highlighting differences.
- Adds changelog story

I'm imagining both Documentation and Changelog stories to become standard for all components.

### DIFF
--- a/libs/spark/stories/navbar.stories.tsx
+++ b/libs/spark/stories/navbar.stories.tsx
@@ -26,13 +26,14 @@ export default {
 } as Meta;
 
 const BluePrendaMonogram = styled(PrendaMonogram)(({ theme }) => ({
-  fontSize: '62px',
+  fontSize: '72px',
   fill: theme.palette.brand.blue,
 }));
 
 const CustomToolbar = withStyles({
   root: {
     gap: 8,
+    height: 80,
   },
 })(Toolbar);
 
@@ -44,13 +45,6 @@ const InboxNavBarButton = withStyles({
     marginRight: 2,
   },
 })(NavBarButton);
-
-const CustomAvatar = withStyles({
-  root: {
-    width: 38,
-    height: 38,
-  },
-})(Avatar);
 
 const Template: Story<NavBarProps> = (args) => (
   <NavBar {...args}>
@@ -78,7 +72,7 @@ const Template: Story<NavBarProps> = (args) => (
       >
         0
       </InboxNavBarButton>
-      <CustomAvatar src="/img/student-boy-2.png" alt="john doe" />
+      <Avatar src="/img/student-boy-2.png" alt="john doe" size="medium" />
     </CustomToolbar>
   </NavBar>
 );
--- a/libs/spark/stories/navbar.stories.tsx
+++ b/libs/spark/stories/navbar.stories.tsx
@@ -5,7 +5,6 @@ import {
   HomeDuotone,
   InboxFilledDuotone,
   MountainDuotone,
-  UserDuotone,
   UsersDuotone,
   PrendaMonogram,
 } from '@prenda/spark-icons';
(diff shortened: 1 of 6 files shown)


seigler (commented):
This is wonderful, but I'm not sure a separate Documentation story is necessary. The Docs view (sibling to the Canvas view in Storybook) shows JSDoc info, which is currently coming via inheritance from Mui (defined [here](https://github.com/mui-org/material-ui/blob/next/packages/material-ui/src/Avatar/Avatar.d.ts) and copied in a build process into [here](https://github.com/mui-org/material-ui/blob/next/packages/material-ui/src/Avatar/Avatar.js#L189)). If we add a new property or change an existing one can't we use the same mechanism and rely on this (mostly automatic) documentation method?

WilliamKelley:
@seigler That's true the information is available there. But there's shortcomings

It provides no links to the
  - underlying component
  - original Props API
  - original CSS API

There's no way for us to audit what the Docs auto-gen picks up vs doesn't pick up -- at least not that I'm aware of -- so we can trust it'll pick up everything, but make me nervous. Additionally, I think there's developer ease in knowing "okay, I'm familiar with the underlying Mui component, and this is exactly how spark customizes the props signature", versus manually comparing the Docs auto-gen to the default Mui API. So I'd rather devs always have a way to get to the original source. The docs add-on also doesn't enumerate the custom classes we provide, so either way we would have to record those.

So with us doing all that, I figure the props isn't that much of an additional step.

What do you think?

WilliamKelley:
Additionally, some of the components aren't getting their props auto-doc'd. It may be only components we don't have a custom wrapper. I think that since all Mui components are wrapped in `withStyles` (`WithStyles` HOC) in the package, it screws with the plugin.

libs/spark/stories/navbar.stories.tsx
```diff
@@ -76,9 +72,7 @@ const Template: Story<NavBarProps> = (args) => (
       >
         0
       </InboxNavBarButton>
-      <UserMenu>
-        <UserDuotone color="onLight" fontSize="large" />
-      </UserMenu>
+      <Avatar src="/img/student-boy-2.png" alt="john doe" size="medium" />
```
shadoath, line 75: Where is this src referencing? Does it need to be added?

WilliamKelley: Assets under the `/libs/spark/public` folder. Webpack loads those files at static assets by default. You can see them working in the storybook.

shadoath (approved):

seigler (approved):
Given those serious deficiencies of the existing docs system, I'm back on board with a documentation page in Storybook. It might be worth examining their MDX documentation patterns, especially [CSF stories with MDX docs](https://github.com/storybookjs/storybook/blob/next/addons/docs/docs/recipes.md#csf-stories-with-mdx-docs)

WilliamKelley:
> 
> 
> Given those serious deficiencies of the existing docs system, I'm back on board with a documentation page in Storybook. It might be worth examining their MDX documentation patterns, especially [CSF stories with MDX docs](https://github.com/storybookjs/storybook/blob/next/addons/docs/docs/recipes.md#csf-stories-with-mdx-docs)

I will make an issue for this, thanks for bringing it up!

PR #215, opened 2021-08-19
feat(Autocomplete): initial support for use-case

### LINKED ISSUE #212: Autocomplete
The practical use-case for a `Menu` with `MenuItem`'s that contain `Checkbox`'s is in the `Autocomplete` component. See here https://material-ui.com/components/autocomplete/#checkboxes. Such a component will also use `Tag` display the selections in the field, instead of Mui's Chip. This pattern is display in the Select Field Figma page: https://www.figma.com/file/y6Ca8BdJtQEd3Of1saRieF/?node-id=949%3A17

This component should be re-exported with validated styling in stories.

For relevant history see
  - designer-filed issue for Menu/MenuItem #54 
  - MenuItem implementation #210 
  - Menu issue separation #211.
  - Tag implementation #203

WilliamKelley:
Closes #212.

Implements the "multiple checkbox select with tag rendering" use-case shown in the 'Select Field" Figma page.

This incorporates a lot of recent work:
 - `Tag`'s for selected option rendering
 - Checkbox items for option rendering
 - `MenuItem` styling ported to `Autocomplete`'s `option` classes slot
   - it was unfortunate to learn it doesn't actually use `MenuItem`'s for its options!

Adds the story for this single use-case.

This really is just the initial support for this component. This component can do _so_ much, but this use-case is the only one outlined so far in Figma.

### DIFF
--- a/libs/spark/CHANGELOG.md
+++ b/libs/spark/CHANGELOG.md
@@ -23,6 +23,9 @@
 
 ### Features
 
+- **Autocomplete**
+  - Added re-export.
+  - Added initial styling support: multiple Checkboxes.
 - **Avatar**
   - Initial implementation.
 - **Divider**
--- a/libs/spark/src/Autocomplete.tsx
+++ b/libs/spark/src/Autocomplete.tsx
@@ -1,0 +1,92 @@
+import * as React from 'react';
+import {
+  Autocomplete,
+  AutocompleteClassKey,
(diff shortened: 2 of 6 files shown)


seigler (approved):
I agree it's too bad Autocomplete doesn't use MenuItem. This is nice.

shadoath (approved):

PR #234, opened 2021-09-08
chore: hook for merging custom and underlying classes

WilliamKelley:
Introduces new hook to DRY out the selective extractions of underlying vs custom classes in our few custom components. The TS signature is as exact as I could make it. One thing that TS had trouble with was correctly inferring the generics for the hook's signature. So, those are explicitly given for each invocation.


libs/spark/src/utils/useTriMergeClasses.ts
```diff
@@ -0,0 +1,46 @@
...
+  underlyingClasses: Partial<ClassNameMap<UnderlyingClassKey>>;
+  customClasses: ClassNameMap<CustomClassKey>;
+} {
+  const underlyingClasses = { ...baseUnderlyingClasses };
+
+  const customClasses = { ...baseCustomClasses };
+
+  // build upon the above initial values
+  for (const [key, value] of Object.entries(passedClasses)) {
+    // a short-circuit type check for whether this key should be merged with
```
seigler, line 30: This explanation is really helpful. Thanks

seigler (approved):

shadoath (approved):

Issue #245, opened 2021-09-09
Abstract DropdownButton


WilliamKelley:
### Missing feature
Ability to use `IconButton` as the anchor element/component of Spark's Dropdown pattern

### Proposal
Generalize `DropdownButton` to `DropdownAnchor` which is an `OverridableComponent` such that it defaults to rendering with `component={Button}` but can be given any of the actionable components: `IconButton` and `AvatarButton` (planned).

The less attractive alternative is adding separate `DropdownIconButton`, `DropdownAvatarButton` components which would just repeat the same simple logic.

PR #246, opened 2021-09-09
feat(Dropdown): remove unneccessary custom components; add placement prop; fix handler swallowing

### LINKED ISSUE #185: Dropdown event handlers
The Dropdown custom component uses its own internal event handling to manage state (open, closed). It might be valuable to emphasize this in the eventual documentation for this component, since the button and menu still accept onClick and onClose props, respectively.
Alternatively, those props could be removed from `SparkDropdownButton` and `SparkDropdownMenu` so overriding them is not possible.

WilliamKelley:
Resolves #185 .

Removes `DropdownDivider`, `DropdownMenuItem` which are equivalent to `Divider`, `MenuItem`. They were relics from before we had those basic components done.

New placement prop for shortcutting the `anchorOrigin` and `transformOrigin`. It follows the Figma file spec.

Extracts `DropdownContext`, `DropdownMenu`, and `DropdownButton` into dedicated files with better utilities/organization.

See future plans in #245


seigler (approved):
:+1:

slingner (approved):
🚀

PR #250, opened 2021-09-17
chore: two-step theme creation for overrides

### LINKED ISSUE #249: Two-step theme creation so overrides can use theme
### Currently
Global component styles overrides are set through the theme. Since all of these override definitions are used in the creation of the theme, they cannot also use the theme object for its other properties like `.palette`, `.typography`, etc.

### Enhancement
Change theme creation to a two-step process where the theme is created _without_ overrides first. Then, create the theme again, this time defining the `overrides` property and explicitly supplying the previous theme to each component's style overrides definition.

This means each component's style overrides should now be a function (when it needs `theme`). Currently, using theme properties in these overrides is a careful process where something like `palette` has to be imported from its deep file (`/styles/palette`) instead of the directory index because that would create a circular dependency.

### Example
```ts
// FILE: src/styles/theme.ts

const initialTheme = createTheme({
  palette,
  props,
  shadows,
  typography,
});

const theme = createTheme({
  ...initialTheme,
  overrides: {
    MuiButton: MuiButtonStyleOverrides(initialTheme),
    ...
  }
});

export default theme;
```

```ts
// FILE: src/Button.ts

export const MuiButtonStyleOverrides = (theme) => ({
  color: theme.palette.text.onLight;
});
```

WilliamKelley:
Resolves #249

### DIFF
--- a/libs/spark/src/Autocomplete.tsx
+++ b/libs/spark/src/Autocomplete.tsx
@@ -1,21 +1,20 @@
 import * as React from 'react';
+import { StyleRules, Theme } from '@material-ui/core';
 import {
   Autocomplete,
   AutocompleteClassKey,
   AutocompleteProps,
 } from '@material-ui/lab';
-import { palette } from './styles/palette';
 import { ChevronDown } from './internal';
-import { typography } from './styles/typography';
-import { StyleRules } from '@material-ui/core';
 
 export default Autocomplete;
 
 export type { AutocompleteProps };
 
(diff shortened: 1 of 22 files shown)


shadoath (approved):
We will continue to use the SparkThemeProvider as is correct?

WilliamKelley:
> We will continue to use the SparkThemeProvider as is correct?

Correct, only the end result is exported, equivalent to before.

seigler (approved):

Issue #242, opened 2021-09-08
Add version number to publish webhook message


WilliamKelley:
Origination: https://prenda.slack.com/archives/C028ZJ179HC/p1629728894000400?thread_ts=1629726770.000100&cid=C028ZJ179HC

WilliamKelley:
Closed in #262

PR #262, opened 2021-09-29
chore: implement common build process for icons and components

### LINKED ISSUE #213: Implement pseudo-tree-shakable builds/bundle for Spark
We achieved a pseudo-tree-shakable bundle for `@prenda/spark-icons` by creating and building with a custom babel config. The config is based off Mui's which preserves `n` icon files with `n` files in the build output, along with `n` type declaration files.

The `@prenda/spark` package does not use this approach though, and the standard Nx build only produces `1` file for `n` components. We should change that.

One trivial part of supporting this is that re-exported components should become the default export of their file. This isn't strictly necessary, but it's the Mui standard and we should follow.

WilliamKelley:
Closes #213 and #242 

- Hoist custom `babel.config.js` and `copy-files.js` to top-level `scripts/` directory. Generalize them slightly to fit both use-cases.
  - dist files are now written to the libraries exact name, not with `-alt` appended.
- `npm-publish` workflow
  - Add descriptive names to each step
  - Add version number (stored in GITHUB_REF env var when action is triggered by `release`)

I've run both build processes locally and they work as expected. I also tried importing a component into a story from the `dist/` directory using a deep import (ex: `import Avatar from '../../../dist/libs/spark/Avatar';` and it both rendered correctly and had correct type and props hinting/annotations.

### DIFF
--- a/.github/workflows/npm-publish.yml
+++ b/.github/workflows/npm-publish.yml
@@ -18,18 +18,28 @@ jobs:
         with:
           node-version: '14'
 
-      - run: npm ci
-      - run: npx nx run spark:test
-      - run: npx nx run spark-e2e:e2e
+      - name: Clean install
+        run: npm ci
+
+      - name: Test
+        run: npx nx run spark:test && npx nx run spark-e2e:e2e
+
       # stories & tests depend on @prenda/spark-icons, but aren't included in build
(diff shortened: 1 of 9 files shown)


slingner (approved):

shadoath (approved):
Nice to see us improving the daily process. Thank you.

PR #282, opened 2021-10-05
chore: replace "Material-UI" with "MUI" and update url's

### LINKED ISSUE #278: Replace outdated links to https://www.material-ui.com
The replacement is `https://v4.mui.com`. Also, "Material-UI" should be replaced with "MUI"

WilliamKelley:
Close #278


shadoath (approved):
And then it will be just MUI

seigler (approved):

PR #287, opened 2021-10-07
chore: improve storybook sidebar hierarchy & SvgIcon story

WilliamKelley:
Close #285 
Close #286 

Aside from those issues, I improved `svgicon.stories.tsx` by using the component directly and using `<path>` tags directly. Why? To have a better output in "Show code" for the story in the Docs tab. (note: the wrapper is still necessary to show the icon colors effectively using different background colors.

Before:

image

After:

image

PR #288, opened 2021-10-07
chore: co-locate stories and tests with components

### LINKED ISSUE #45: Co-locate components, stories, and tests
Currently Spark places component definition files under `src/`, stories under `stories/` and tests under `test/`.

I propose we move to creating directories for each component under `src/` that include all of the definition file, story file, and test file.

Pros:
- Developer doesn't have to jump between directories.
- Easier to verify/ensure that every component has at least an individual story and test file

Cons:
- No clear place for stories that show multiple components in concert
  - Solution: keep `stories/` exactly for this purpose, but still co-locate all component-specific stories.
- Same as above but for tests
  - Solution: Same as above

In terms of a tree diagram for a single component that combines :

Current:
```bash
src
├── elements
│   ├── button.tsx
│   ├── index.ts
stories
├── elements
│   ├── button.stories.tsx
test
├── button.spec.tsx
```

Proposed:
```bash
src
├── elements
│   ├── button.tsx
│   ├── button.stories.tsx
│   ├── button.spec.tsx
│   ├── index.ts
```

Proposed with #38 :
```bash
src
├── Button
│   ├── Button.tsx
│   ├── Button.stories.tsx
│   ├── Button.spec.tsx
│   ├── index.ts
```

WilliamKelley:
Builds off #287 

Close #45 

This is an ugly PR. Lots of files changed, but it's reviewable by commit, but there are many commits... sorry :/

The `src/tests/` directory has been removed after all of its files relocated. However `src/stories/` remains to house stories of `@prenda/spark-icons` components (icons, logos, illustrations) and of `CheckboxGroup` which is a pattern / composition of many components, but not strictly one itself. This is opposed to compositions like `dropdown.stories.tsx` which was renamed to `DropdownContext.stories.tsx`. Though it's longer and perhaps less digestable, it finds an exact home in the `src/DropdownContext/` directory and it's accurate since it's the parent component, required for all dropdowns; same goes for `tabs.stories.tsx  ->  TabContext.stories.tsx`. Following this pattern it deserves to live under `FormControl.stories.tsx` given that's the parent. Just not sure how to best make it as realize-able (not the word i'm looking for, but...) as it is when labeled as "Checkbox Group".

Additionally, the existing `*.spec.*` files of each component were reduced to test just that component, not auxiliary ones. After all, they're just checking truthiness. With time, every component should have that spec, and the actually patterns should be appropriately tested deliberately.

### DIFF
--- a/libs/spark/src/Checkbox/Checkbox.stories.tsx
+++ b/libs/spark/src/Checkbox/Checkbox.stories.tsx
@@ -2,6 +2,7 @@ import * as React from 'react';
 import { Meta, Story } from '@storybook/react/types-6-0';
 import Checkbox from './Checkbox';
 import FormControlLabel from '../FormControlLabel';
+import { DocumentationTemplate } from '../../stories/templates';
 
 export default {
   title: 'PDS/@ps/Checkbox',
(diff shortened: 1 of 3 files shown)


WilliamKelley:
@seigler Thanks! thoughts on my last comment?

seigler:
> @seigler Thanks! thoughts on my last comment?

I agree Checkbox Group belongs under Form Control. To improve discoverability, maybe we can link to that section in the docs for Radio and Checkbox controls?

Issue #292, opened 2021-10-08
Tree-shaking is ruined by theme


WilliamKelley:
## Problem
Since our theme creation leverages custom props and overrides of all components by importing them at evaluation-time (in `src/theme/(props|overrides).ts`), all of those files are included when a consumer uses any of our components. This ruins tree-shaking / pseudo-tree-shaking.

## Solution
Don't leverage theme-level overrides for props and styles -- at least not in the same manner that we do currently. An alternative is 
1. Modify re-export of `withStyles` to be given a `defaultTheme` of our theme (but our theme without props & overrides ofc).
2. Re-export components after invoking `withStyles` around it.
    - it'll be beneficial in this step to give custom theme-level names to all components with a `MuiSpark` prefix so they have completely isolated global "identifiers"

### Side effects
- Consumers won't need to set up `SparkThemeProvider` anymore... 0.o
  - Consumers can use existing components or default Mui components side-by-side without any theme conflicts.

PR #302, opened 2021-10-14
chore: rename file extensions in prep of fixing #292

WilliamKelley:
part of #292 

Unfortunately, renaming and changing a file too drastically in one PR will prevent GH from truly carrying the rename forward even if you split them into separate commits; see #301 (in progress) for evidence. This greatly hurts the reviewability of the PR for files that had anything substantial in them before a rename and altering. These are all the files I anticipate needing to rename for making changes (the index.js => .ts aren't a _need_, but they should've been this way from the start).


WilliamKelley:
debating if i want to move forward on this path

PR #305, opened 2021-10-14
fix: isolate theme overrides and props from component exports 

### LINKED ISSUE #292: Tree-shaking is ruined by theme

WilliamKelley:
Close #292 

This isn't the ideal solution of "support component stylings / default props without being under theme context", but it is a solution separate from that lofty endeavor. `...DefaultProps` and `...StyleOverrides` are moved out of the file that the component is exported from and into `defaults.ts` under the same directory. For every instance, I've added better and more consistent typings. The type for `...DefaultProps` will prevent invalid prop keys or values from being given, and the type for `...StyleOverrides` will prevent invalid class keys (rule names) or property values from being given. The `theme/(props|overrides).ts` files import directly from these files and thus all the associated components are not evaluated part of a bundle that just uses `theme`.


WilliamKelley:
I've separated the "ideal" solution from #292 into a feature discussion #307, so that we can effectively close #292 without losing information / motivation.

WilliamKelley:
Haha i get ya. That's why we have Chromatic to tell us the story of if anything changed!

PR #309, opened 2021-10-19
feat(Stepper): initial implementations; tests; stories

WilliamKelley:
This is a large PR, but reviewable by commit.

This component is called "Steps" in the [Figma spec](https://www.figma.com/file/y6Ca8BdJtQEd3Of1saRieF/DS%3A-Web-Components?node-id=3295%3A4725). It's not finished, so a corresponding issue was never filed, however Code Club needs this for a new product design, and I believe that the implementation is well-defined in terms of styling and in terms of API by extending or re-implementing [MUI's Stepper](https://v4.mui.com/components/steppers/#stepper)

This design covers implementing seven components:
- `StepConnector`: 
  - the line between each step, extending vertically or horizontally (`orientation`).
  - meant to appear as a child of `Step`
  - not meant to be commonly used by consumer, but available
- `StepIcon`
  - the circle with text (number) that resembles an `IconButton`.
  - meant to appear as a child of `StepLabel`
  - not meant to be commonly used by consumer, but available
- `StepLabel`
  - a composition of `StepIcon` beside a label -- that label can be positioned by `orientation` or alternatively.
  - meant to appear as a child of `Step` or `StepButton`.
- `StepContent`
  - provides a "description" for an `active` or `expanded` step in a vertical `orientation`. 
  - meant to appear as a sibling of `StepLabel` or `StepButton` and as a child of `Step`
- `StepButton`
  - wraps `ButtonBase` around it's `StepLabel` child (if the child is not `StepLabel`, it detects that and wraps whatever it is in `StepLabel`)
  - meant to appear as a child of `Step`
- `Step`
  - a parent for all of the above components that takes care of most of their props.
  - meant to appear as a child of `Stepper`
- `Stepper`
  - a parent for `Step`'s that takes care of its props
  - the only component whose props are really meant to all be exposed to the consumer.

All of the components have stories _and_ tests. 
- Stories cover all design variations I deemed reasonable. I tried a new "storyName" convention where prop names are lowercase, pseudo-state/class names are lowercase with a colon (":") prefix, and only descriptive words are uppercase (like Configurable, Labels, Default, etc.) -- there was doubling of words given the `active` prop and `:active` pseudo-class that initially inspired this pattern, but I believe it has reason outside of that problem solution. 
- Tests are _not_ superficial and try test the output not the internals (guiding principles of [testing-library](https://testing-library.com/docs/react-testing-library/intro)). When testing a re-exported component, I didn't want to test everything Mui does since we can assume that's guaranteed. But when re-implementing, I tried to carry over reasonable tests.

A few of the basic ones were re-implemented according to MUI's source because the type-safe customization was going to be as long or longer than the original source file, and there wasn't much complexity that the MUI source was abstracting.

The rest were re-exported with custom styling and some re-typing for our Storybook. Now, the re-typing is a bit worrisome since most of the components aren't meant to have their props exposed to the _user_, only to it's intended parent component. That's why MUI didn't document them, but then creating stories for all of their states individually _and_ specifying their props in tests would require re-typing in some form. Up for feedback on the best way to handle that. I will note that we don't encounter these complexities with things like `TabList`, `TabPanel` because those use `TabContext` internally. I think these components were dev'd before Context was popular / stable and haven't gotten re-vamped because there's no real a11y concerns with them?

To aid in the various implementations, I added a new util, `useMergeClasses`, that is simpler and suitably more succinct for when the custom class key and underlying class key are the same, or for in re-implementations when our class key is the base class key.

Lastly, I made a couple small changes to `SvgIcon`:
- widening the ref calm some type-errors.
- adding a `data-testid` attribute by default that we can query for in tests. This is what MUI does. I also added a babel plugin to our custom build process to strip these attributes at build time.

### DIFF
--- a/libs/spark/CHANGELOG.md
+++ b/libs/spark/CHANGELOG.md
@@ -2,8 +2,31 @@
 
 # [vNext](https://github.com/prenda-school/prenda-spark/compare/v0.14.0...vNext) (yyyy-mm-dd)
 
+### Fixes
+
+- **SvgIcon**
+  - Widened `ref` type to `any`.
+
 ### Features
 
+- **Step**
+  - Re-exported.
+  - Customized styling.
+- **StepButton**
+  - Initial implementation.
+- **StepConnector**
+  - Initial implementation.
+- **StepContent**
+  - Re-exported.
+  - Customized styling.
+- **StepIcon**
+  - Initial implementation.
+- **StepLabel**
+  - Re-exported.
+  - Customized styling.
+- **Stepper**
+  - Re-exported.
+  - Customized styling.
 - **Typography**
   - New prop: `lowContrast`. When given, applies the "low contrast" version of the current `color` -- currently defined for use with `color="light"` and `color="dark"`.
 - **makeStyles**


slingner:
> Really nice.

Agreed!

PR #319, opened 2021-10-29
chore: improve husky hooks and CI workflow

WilliamKelley:
Removes testing and linting on both Husky pre-commit and pre-push hook scripts, and improves the format command.

Why remove testing and linting? Both of these processes take an amount of time that reduces developer productivity and increases frustration (speaking from my experience and multiple peers). So, I've moved those checks into the Continuous Integration workflow, increasing its value and utility for contributors (imo, it had low value before; I knew it would pass because I couldn't have pushed it up otherwise).

We don't need to be worried about "every commit should be linted and tested". First, we squash and merge, so it's not important beyond the PR state. Second, in the PR stage, there's good reason to *want* to commit or push failing code: so others can pull it down and collaborate to fix failures, or just let the CI infrastructure run the tests instead of your machine; of course, `--no-verify` is always an option, but we should be in the business of choosing the best and sane defaults. The CI workflow will now fail when linting, formatting, or testing has a non-zero exit code -- linting was adjusted to error on warnings.

Finally, the existing command to format had a flaw imo; it would write its formatting changes, but wouldn't stage them in the commit, so you'd either have to amend the commit, or include it in your next one; blegh. Instead, I customized the command to eliminate the barrier of having a commit there but keep the auto-write and fail the hook, letting the user know exactly what to do for the hook to succeed -- that should improve DX.

Looking forward to your feedback.

### DIFF
--- a/.github/workflows/chromatic.yml
+++ b/.github/workflows/chromatic.yml
@@ -1,6 +1,6 @@
 name: 'Chromatic'
 
-on:  
+on:
   push:
     branches: [main]
   pull_request:
@@ -10,9 +10,9 @@ jobs:
   chromatic-deployment:
     if: github.event.pull_request.draft == false
     runs-on: ubuntu-latest
-    steps: 
-      - uses: actions/checkout@v1
-        
+    steps:
+      - uses: actions/checkout@v2
+
       - name: Install Dependencies
         run: npm i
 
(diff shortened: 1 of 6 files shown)


WilliamKelley:
Really not sure why the CI is failing. From what I've been able to google, the v2 checkout action with `fetch-depth: 0` should have solved the `main` ref not being available... help appreciated

WilliamKelley:
Passing now, i just removed the affected lint to be an "everything" lint

seigler (approved):
I am excited about this one

shadoath (approved):

PR #321, opened 2021-11-01
chore(storybook): add branding

WilliamKelley:
Replaces default storybook branding with Prenda Design System. I created this "logo" independently, with the intention filling the 350x150 constraint as reasonably as possible.

Turns off [sidebar "roots"](https://storybook.js.org/docs/react/configure/sidebar-and-urls#roots) since it's now redundant with the branding.

image

The brand logo svg is added to both the assets served for the storybook (`libs/spark/public/*`) and the repository root's public assets (`public/*`) in anticipation of use in the rewrite of the root `README` when the name change from Spark to PDS is completed.


WilliamKelley:
Pretty sure all Chromatic snapshots will be reset given the new paths, so that can be disregarded.

shadoath (approved):
Nice! I like that we can add this to storybook.

seigler (approved):

PR #326, opened 2021-11-11
feat(Collapse): backport v5 changes for horizontal orientation

WilliamKelley:
The `<Collapse>` from MUI v4 only supports vertical orientation, but v5 supports horizontal as well.I'd like to have the functionality immediately for a consuming project, so I essentially copy-pasted the file from right before v5's migration to emotion, added types in the file and adjusted it to our makeStyles / useMergeClasses paradigm.

### DIFF
--- a/libs/spark/src/Collapse/Collapse.stories.tsx
+++ b/libs/spark/src/Collapse/Collapse.stories.tsx
@@ -28,6 +28,32 @@ export default {
   },
 } as Meta;
 
+const useIconStyles = makeStyles((theme) => ({
+  paper: {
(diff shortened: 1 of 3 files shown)


shadoath (approved):

WilliamKelley:
I would migrate MUI's test as well, but that's non trivial since they have a custom `render` function that's not exposed to consumers.

WilliamKelley:
Re-exported props.

Added stories to validate all cases. As good as tests for now.

WilliamKelley:
@seigler @slingner bump

slingner (approved):

PR #327, opened 2021-11-13
feat(Link): re-exported with custom styles, props

WilliamKelley:
Backport of a PDS v2 component spec using the best palette parallels. This is immediately useful.

Figma spec page: https://www.figma.com/file/6MR4jvUO9jG5pKEogHszHR/?node-id=4284%3A5154

Between "inline" and "standalone", I'm assuming that "inline" will be the more common use-case, so `standalone` is an opt-in prop -- replaces underlying `underline` prop.

All stories spec'd out.


shadoath (approved):

WilliamKelley:
@seigler @slingner bump

slingner (approved):

PR #329, opened 2021-11-13
feat(Button): customize without `ThemeProvider`

WilliamKelley:
Moves prop and style defaults into re-export.

Chromatic shows minor styling changes, but it's actually more correct now. The difference was that medium buttons of different variants were different heights before... dunno how that got through.


shadoath (approved):

WilliamKelley:
Closing, see #337

PR #331, opened 2021-11-13
feat(IconButton): customize without `ThemeProvider`

WilliamKelley:
Moves default props to component, adjusts test. Minor, unrelated corrections in the stories.

PR #332, opened 2021-11-13
feat(Checkbox): customize without `ThemeProvider`

WilliamKelley:
Builds on #331 since internal composition goes: Checkbox -> SwitchBase -> IconButton -> ButtonBase

Moves default props and styles to component file, adjusts test. Minor, unrelated corrections in the stories.

The component file itself takes a different approach than usual. Since MUI's implementation is quite basic, only containing styling specifics and icon ternaries (https://github.com/mui-org/material-ui/blob/v4.x/packages/material-ui/src/Checkbox/Checkbox.js), the implementation is mocked. All the important logic is contained within their `SwitchBase` implementation, which we don't re-implement.

At first I thought this increases maintenance cost, but really I'm not sure it does. It's only sure that initial implementation cost is higher, but now it's done. There's less props and classes layers to worry about the performance or complexity of now.


WilliamKelley:
Converting to draft because of problem discovered in #333

WilliamKelley:
Close in favor of #335

PR #340, opened 2021-11-17
chore(stories): cleanup & normalize behaviors

WilliamKelley:
Bunch of chores for story files

- Change `import { Meta ...` to `import type { Meta ...`
- Import from `/src` index (`... from '..';`) to simplify imports (storybook/webpack tree-shakes and this is cleared from bundling)
- Correct `: Story` typing; exported stories should receive the type, not non-exported templates.
- Remove unnecessary re-templating of `ChangelogTemplate` and `DocumentationTemplate`.
- Change re-typing variable names `Typed<ComponentName>` -> `Sb<ComponentName>` (short for "Storybook", the prefix we already use for Storybook-custom docs props.
- Where Storybook's auto-docs doesn't pick up props or their descriptions, re-declare props and re-use original definitions where possible (descriptions are also picked up through this reuse). Also, remove excess ripple props where possible.
- Change "Storybook-only" comment format.
- Change some "Storybook-only" prop name formats; where applicable, the name references the component it's passed to if it's not the component to which the story belongs to (ex: `sb_TabList_onChange` because it's under `TabContext`).
- When a prop is of type `React.ReactNode`, implement a custom mapping in the `Meta['argTypes']`, and eliminate conditional logic in template renders.

### DIFF
--- a/libs/spark/src/Autocomplete/Autocomplete.stories.tsx
+++ b/libs/spark/src/Autocomplete/Autocomplete.stories.tsx
@@ -1,11 +1,13 @@
 import * as React from 'react';
-import { Meta, Story } from '@storybook/react/types-6-0';
-import Autocomplete from './Autocomplete';
-import Checkbox from '../Checkbox';
-import ListItemText from '../ListItemText';
-import ListItemIcon from '../ListItemIcon';
-import TextField from '../TextField';
-import Tag from '../Tag';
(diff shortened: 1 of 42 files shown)


slingner (approved):

seigler (approved):

seigler:
There was a lot here. Great work

PR #347, opened 2021-11-19
feat(__next__palette): PDS v2 palette

### LINKED ISSUE #322: Updated palette
## Color changes

We are updating the Prenda secondary colors and adding range to the color system to support illustration work and to have colors that identify with the Prenda brand more effectively. 

These changes are a part of a bigger effort to control/stabilize the brand. 

[Figma documentation](https://www.figma.com/file/w9lodWeivyS4GHdBNIYaJr/DS%3A-Foundations?node-id=1%3A2
)

WilliamKelley:
Closes #322

Implements the new palette on a very-clearly experimental field on theme: `theme.__next__palette`. The idea is that when all internal dependencies on `theme.palette` are removed, then the `__next__` will take over. In the meantime, consumers who need this PDS v2 palette will have experimental access to it -- avoiding having to create their own sources of truth.

I reference `globalTokens` and `aliasTokens` based on my reading of Adobe Spectrum's Foundation explainer on design tokens: https://spectrum.adobe.com/page/design-tokens/ . I recommend reading every item under "Foundation", but a particularly enlightening one is International design - Colors https://spectrum.adobe.com/page/international-design/#Colors . I believe that from this, you can see where `aliasTokens` come in. They're composed of (probably) global tokens, that are indisputably correct (red is red to everyone), but are a named abstraction (error is error to everyone, but different colors to some). From this, we get `background` and `text` on our palette as functional aliases. I've described them as "experimental" in their respective stories, because they aren't officially reflected in the Figma spec, and are kind-of ad-hoc for now.

In the changelog, I've detailed the "Planned Migration" that I have confidence will work for our consumers. Feedback appreciated.

Also if you could take the time to tediously double-check I've typed in all the hex values correctly, that'd be great.

### DIFF
--- a/libs/spark/CHANGELOG.md
+++ b/libs/spark/CHANGELOG.md
@@ -78,12 +78,12 @@ _This section details previous of breaking changes or experimental features and 
       - `palette.info.main` -> removed (\_note: if you require this alias, use `palette.blue[600]`)
       - `palette.success.main` -> removed (\_note: if you require this alias, use `palette.green[600]`)
     - "neutrals":
-      - `palette.common.white` -> `palette.neutrals[0]`
-      - `palette.grey.lighter` -> `palette.neutrals[60]`
-      - `palette.grey.light` -> `palette.neutrals[70]`
-      - `palette.grey.medium` -> `palette.neutrals[80]`
-      - `palette.grey.dark` -> `palette.neutrals[90]`
-      - `palette.blue[5]` -> `palette.neutrals[600]`
+      - `palette.common.white` -> `palette.neutral[0]`
+      - `palette.grey.lighter` -> `palette.neutral[60]`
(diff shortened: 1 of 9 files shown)


WilliamKelley:
The only thing I'm not confident about is if the brand background should dictate it's own text keys, like inverse does.

It's reasonable to constrain default and alternative to have the same text color, and inverse obviously must have its own, but putting the constraint that brand must be the same text color as inverse seems a bit much. But the alternative is adding brand-prefixed text keys, which i don't love since there are already so many, but maybe that just means it's reasonable...

WilliamKelley:
Converting from draft to see stories in Chromatic

WilliamKelley:
Ready for review.

shadoath (approved):
Good writeup

seigler (approved):
That changelog looks like it's halfway to being a codemod. Nice.

PR #351, opened 2021-11-22
feat(__next__typography): PDS v2 typography

WilliamKelley:
_warning: the (disk) size of this PR is very large, just under 30 MB, because of the complete storage of fonts and their source zip files. This also affected the diff -- I recommend excluding files with extensions `(woff|woff2|zip|txt)` from the diff / changeset._

Implements PDS v2 typography. Takes the same approach as in #347 where the tokens are attached via `__next__typography`.

- `assets/` 
  - New directory for complete storage of converted font files and their original source downloads.
  - Useful for when a new weight is needed, or if a consumer needs a weight outside our spec. Files from different sources can be for different revisions, or be of varying sizes, maybe having varying appearance, etc. And complete re-conversion is a multi-step process. This seems good to check in, even though it's large.
- `../src/theme/__next__typography`
  - Added story
    - When stable, will take role of the "Overview" Typography story, but lie under `theme/` stories dir.
- `../src/theme/__next__fontFaces`
  - Added story to explain the process behind font files
    - Provides direction to copy the contents of the `../public/fonts` dir into the consumer's own app (finally) (this should be extracted into some sort of "Getting started" story in the future).
- `../public/internal/`
  - New directory for storage of story asset files that aren't needed to achieve implementation specs.
  - Triggered by usage/showcase of font weights in "Foundations" figma files that aren't spec'd typographic variants. Thus, consumers don't need the added weight of these files.
- 	`../public/fonts/(inter|poppins)/{weight}.woff(2)`
  - "directory-ized" font file organization. Shortened filenames to just the weight, given that their family is obvious (dir name) and "web" / "webfont" is superfluous.

### TODO's
There are six outstanding `:TODO:` comments, for which feedback has been requested: https://github.com/prenda-school/design-system/issues/350#issuecomment-978044953 . However, the PR is reviewable without full address. I will ensure these are addressed before merging.


seigler (approved):
> "directory-ized" font file organization

Yes!
Nice.

shadoath (approved):

slingner (approved):

WilliamKelley:
Closed outstanding TODO's. 

Changes:

- `assets/` 
  - added `robotomono-*` files
- `../src/theme/__next__typography`
  - Added `code` variant
  - Add variant in story, replaced use of Source Code Pro to Roboto Mono
- `../src/theme/__next__fontFaces`
  - Added Roboto Mono regular/italic font faces.
  - Added story for monospaced font - roboto mono
- `../public/internal/`
  - Removed duped files
  - Added all weights of robotomono
  - Removed `.woff` extension files since those are only supplied for IE compatibility, and design system dev env's don't need this.
- 	`../public/fonts/(inter|poppins|robotomono)/{weight}.woff(2)`

seigler (approved):
That's a long license

PR #363, opened 2022-03-16
First pass at adding 'destructive' variant to buttons

shadoath:
Currently getting : 
```
Warning: Failed prop type: Invalid prop `variant` of value `destructive` supplied to `ForwardRef(Button)`, expected one of ["contained","outlined","text"].
```

### DIFF
--- a/libs/spark/src/Button/Button.stories.tsx
+++ b/libs/spark/src/Button/Button.stories.tsx
@@ -119,6 +119,23 @@ const VariantAndSizeTemplate = (args) => (
         Label
       </Button>
     </span>
+
+    <span>Destructive</span>
+    <span>
+      <Button {...args} variant="destructive" size="large">
+        Label
+      </Button>
+    </span>
+    <span>
+      <Button {...args} variant="destructive" size="medium">
+        Label
+      </Button>
+    </span>
+    <span>
+      <Button {...args} variant="destructive" size="small">
+        Label
+      </Button>
+    </span>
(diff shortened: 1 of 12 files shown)


WilliamKelley:
This is a MUI v4 limitation. In order to declare a custom variant value, you would need to create a custom wrapper that captures it and doesn't pass it to the base MUI component. Otherwise MUI will always error.

Currently Button is just a re-export and customized by its theme, so you have to adhere _exactly_ to its base API.

To do the custom capturing, you can look at something like the Tag component as an example.

---

Also, all PDS v2 conversion should happen under a separate component, that starts as a duplication, is prefixed with __next__, etc.

shadoath:
>Also, all PDS v2 conversion should happen under a separate component, that starts as a duplication, is prefixed with next, etc.

Should I do the whole folder as `__next__button` or each file in the button folder?

WilliamKelley:
`__next__Button` and dupe the folder.

WilliamKelley:
Also, it might be more productive to start with changing the existing variant designs first, but leaving the variant names. Then doing a custom wrapper with the actual, custom variant names and the new destructive variant. They'd be separate PRs. Otherwise you might get more trouble than its worth from skipping that.

shadoath:
I'm thinking to close this an start again with the changes you have made removing __next__.

WilliamKelley:
@shadoath Sure, up to you.

PR #366, opened 2022-03-18
feat(Unstable_Typography): initial implementation

### LINKED ISSUE #365: Adapt styles based on props, instead of custom classes capture
When implementing custom variants or other prop values that need a class name, we have to do a pretty complicated capturing of classes, verbose class name determination, etc.

I propose that we bite the bullet on [adapting styles to props](https://v4.mui.com/styles/basics/#adapting-based-on-props). I avoided this because of runtime performance concerns and because v5 did not seem like it was moving it that direction at the time, however v5 _fully_ embraces this as the only option.

It makes our jobs easier as we can get rid of some boilerplate, and our consumers won't have to deal with ugly css class specificity workarounds (i.e. `'&&&'`) to override our overlapping determinations.

I'll pioneer this with my work on Unstable_Typography.

WilliamKelley:
Closes #365 by showing viable implementation of adapting styles based on props. (@seigler might be of interest to you since I shot this down when you did it way back when).

New component for preview of PDS v2 Typography. Adds semibold font file (would have been bold but that was already mistakenly added). Adds changelog entry with Planned Migration from existing Typography. Adds test to show it doesn't require the package's `theme` being in a parent `ThemeProvider`.

Also:

- fixes some old changelog mistakes
- fixes font face declaration of misnaming semibold weight font face declaration as bold.


WilliamKelley:
Also on the "Unstable_" prefix. Eslint still warns, so meh to that. I would like to incorporate "Preview" or "V2" in there somehow without being annoying or confusing

- "UnstablePreview_Typography" seems too long, idk
- "UnstableV2_Typography" seems confusing -- i'm talking about PDS v2, but a consumer would think package version v2.
    - One option is we can release what we have now as v1 stable, and then this prefix makes sense?

or im just wavering and nitpicky

libs/spark/src/Unstable_Typography/Unstable_Typography.stories.tsx
```diff
@@ -0,0 +1,162 @@
...
+    <VariantSection
+      variant="T14"
+      details={['14px/20px/4%/uppercase', 'Extrabold', 'Poppins']}
+    >
+      Passion to learn
+    </VariantSection>
+    <VariantSection
+      variant="body"
+      details={['16px/24px', 'Regular', 'Inter', '(cv05,ss03)']}
+    >
```
shadoath, line 134: Nice, I like how you did this.

libs/spark/src/Unstable_Typography/Unstable_Typography.tsx
```diff
@@ -0,0 +1,162 @@
...
+    root: (props: Unstable_TypographyProps) => ({
+      ...(props.variant === 'display' && {
+        ...theme.unstable_typography.display,
+        color: theme.unstable_palette.text.heading,
+        ...(props.color === 'inverse' && {
+          color: theme.unstable_palette.text.inverseHeading,
+        }),
+      }),
+      ...(props.variant === 'T32' && {
+        ...theme.unstable_typography['T32'],
```
shadoath, line 44: Maybe this is naive but could you do this:
```suggestion
      ...{theme.unstable_typography[props.variant ],
```

My guess is that does not allow for good type-checking and if a value was passed in that did not match it would blow up.

libs/spark/src/theme/unstable_fontFaces.ts
```diff
@@ -88,19 +88,35 @@ const interItalic: CSS.AtRule.FontFace = {
   ].join(', '),
 };
 
-const interBold: CSS.AtRule.FontFace = {
+const interSemibold: CSS.AtRule.FontFace = {
```
shadoath, line 91: Good correction here.

shadoath (approved):

libs/spark/src/Unstable_Typography/Unstable_Typography.stories.tsx
```diff
@@ -0,0 +1,162 @@
...
+import { capitalize } from '../utils';
+
+export const SbUnstable_Typography = (props: Unstable_TypographyProps) => (
+  <Unstable_Typography {...props} />
+);
+
+export default {
+  title: '@ps/Unstable_Typography',
+  component: SbUnstable_Typography,
+  excludeStories: ['SbUnstable_Typography'],
```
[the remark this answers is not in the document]
WilliamKelley: it means "Storybook" (not Skylar Bolton ;) ). The purpose of it is that there's an addon to auto-populate the args/controls storybook addon based on the typing of the component props. The addon only works with _named_ exports, but all of our components are _default_ exports (for another necessary reason). So this naming is a "scoping" of the component so the addon can read it.

libs/spark/src/Unstable_Typography/Unstable_Typography.stories.tsx
WilliamKelley: thanks, the previous way was SO confusing to look back on

libs/spark/src/Unstable_Typography/Unstable_Typography.tsx
WilliamKelley: You could actually, but yes the typing would complain. Also we need it anyways for variant-specific colors.

slingner (approved):
Great work on this, William!

PR #368, opened 2022-03-18
fix(theme): make unstable theme properties available on default theme

WilliamKelley:
Fixes `unstable_palette` and `unstable_typography` not being available on the default theme set for the styling utilities. That theme is the `initialTheme` variable, not the final `theme` variable because the latter includes props and classes default that the utilities are required for (circular dep).

The end result is that consumers will be able to access these properties through styling utilities **without** having to render a `ThemeProvider` as an ancestor.


shadoath (approved):
Yes! this is a great addition

slingner (approved):

PR #370, opened 2022-03-28
fix: style :focus-visible instead of :focus

### LINKED ISSUE #354: Target :focus-visible instead of :focus
Read about it https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-visible

`:focus` has the pitfall of applying after a mouse action is taken, where as `:focus-visible` does not. The latter will _only_ display on keyboard navigation. I believe this is the original design intent of our focus styles. Mouse interactions are covered by `:hover` and maybe `:active`.

WilliamKelley:
Close #354

### DIFF
--- a/libs/spark/src/Banner/Banner.stories.tsx
+++ b/libs/spark/src/Banner/Banner.stories.tsx
@@ -76,7 +76,7 @@ SeverityCloseFocus.args = {
     return;
   },
 };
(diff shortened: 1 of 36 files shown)


shadoath (approved):
Whew, I hope you could do this with a find and replace.

slingner (approved):
💯

Issue #350, opened 2021-11-19
New design system components [mega issue]


alecdanz:
## Simple components 
Simple components are thought of as components that just reference foundational design system elements such as colors and type. 

- [ ] Avatar – [Spec](https://www.figma.com/file/y6Ca8BdJtQEd3Of1saRieF/DS%3A-Web-Components?node-id=949%3A465)
- [ ] Breadcrumb – [Spec](https://www.figma.com/file/y6Ca8BdJtQEd3Of1saRieF/DS%3A-Web-Components?node-id=949%3A0)
- [ ] Button – [Spec](https://www.figma.com/file/y6Ca8BdJtQEd3Of1saRieF/DS%3A-Web-Components?node-id=29%3A0)
- [ ] Checkbox – [Spec](https://www.figma.com/file/y6Ca8BdJtQEd3Of1saRieF/DS%3A-Web-Components?node-id=1056%3A1286)
- [ ] Date picker – [Spec](https://www.figma.com/file/y6Ca8BdJtQEd3Of1saRieF/DS%3A-Web-Components?node-id=954%3A721)
- [ ] Flag – [Spec](https://www.figma.com/file/y6Ca8BdJtQEd3Of1saRieF/DS%3A-Web-Components?node-id=2379%3A4207)
- [ ] Input – [Spec](https://www.figma.com/file/y6Ca8BdJtQEd3Of1saRieF/DS%3A-Web-Components?node-id=207%3A0)
- [ ] Link – [Spec](https://www.figma.com/file/y6Ca8BdJtQEd3Of1saRieF/DS%3A-Web-Components?node-id=4284%3A5154)
- [ ] Pagination – [Spec](https://www.figma.com/file/y6Ca8BdJtQEd3Of1saRieF/DS%3A-Web-Components?node-id=949%3A463)
- [ ] Radio – [Spec](https://www.figma.com/file/y6Ca8BdJtQEd3Of1saRieF/DS%3A-Web-Components?node-id=1056%3A1285)
- [ ] Steps – [Spec](https://www.figma.com/file/y6Ca8BdJtQEd3Of1saRieF/DS%3A-Web-Components?node-id=949%3A462)
- [ ] Switch – [Spec](https://www.figma.com/file/y6Ca8BdJtQEd3Of1saRieF/DS%3A-Web-Components?node-id=2129%3A4193)
[...]

PR #371, opened 2022-03-28
feat(Unstable_Link): initial implementation

WilliamKelley:
Implements PDS v2 Link. Part of #350.


libs/spark/src/Unstable_Link/Unstable_Link.tsx
```diff
@@ -0,0 +1,94 @@
...
+      standalone,
+      ...other
+    } = props;
+
+    const classes = useStyles({ color, standalone });
+
+    return (
+      <MuiLink
+        classes={{
+          root: clsx(classes.root, classesProp?.root),
```
shadoath, line 84: Does this mean the classes passed in would need to have the `root` value?

shadoath (approved):

libs/spark/src/Unstable_Link/Unstable_Link.tsx
WilliamKelley: yes, and only the root key. no other. That's only if the consumer wants to use the classes API

slingner (approved):

PR #374, opened 2022-03-29
feat(Unstable_SvgIcon): initial implementation

WilliamKelley:
Implements PDS v2 SvgIcon. Part of https://github.com/prenda-school/design-system/issues/350.

Note that this component is not spec'd by the "components" Figma file (to my displeasure). Font sizes are extrapolated from ["icons" file, "Cover" page](https://www.figma.com/file/BNkemx3eS6ayFVn4mij6tO/?node-id=410%3A50). Colors are extrapolated from color palette comments (see screenshot below); `"normal"` was used instead of `"default"` since the default is actually `"inherit"`, so that would have been a misnomer. As in the previous version of the component. `"inherit"` is the default for font size and color because it makes styling in components **much** easier. Also, icons are used predominantly in, not outside of, components that consider them in styling.

image


libs/spark/CHANGELOG.md
```diff
@@ -2,6 +2,40 @@
...
+- **Unstable_SvgIcon**
+  - Initial implementation of `SvgIcon` replacement according to PDS v2.
+  - Supports rendering _without_ `theme` being in an ancestor `ThemeProvider`.
+  - CSS API Changes:
+    - Removed all class keys _except `"root"`_.
+  - Props API Changes:
+    - `classes`
+      - Removed all class keys _except `"root"`_.
+    - `color`
+      - Removed values: `"dark"`, `"light"`, `"white"`, `"disabled"`, `"error"`, `"success"`, `"warning"`, `"info"`.
```
shadoath, line 18: With the error/success/warning/info removed, is there a way to use these in notices that have these settings?

libs/spark/src/Unstable_SvgIcon/Unstable_SvgIcon.stories.tsx
```diff
@@ -0,0 +1,67 @@
...
+  </Unstable_SvgIcon>
+);
+
+export const Playground: Story = Template.bind({});
+
+export const Default: Story = Template.bind({});
```
shadoath, line 26: What does `Template.bind({})` do? what might be passed as params?

libs/spark/CHANGELOG.md
shadoath: or do we now have to be explicit?

shadoath: Or from what we did today it will inherit what color it needs to be, right?

shadoath (approved):
What

libs/spark/src/Unstable_SvgIcon/Unstable_SvgIcon.stories.tsx
WilliamKelley: It lets you reuse the template, but have the controls panels be separate for every story, so they don't all affect each other.

WilliamKelley:
> What

😅 yes?

slingner (approved):

libs/spark/src/Unstable_SvgIcon/Unstable_SvgIcon.stories.tsx
WilliamKelley: Also I dont think any params can be passed. it's like [`Function.prototype.bind`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_objects/Function/bind) I think

WilliamKelley: resolving for merge. feel free to elicit more discussion though

PR #375, opened 2022-03-30
feat(Unstable_Button): initial implementation

shadoath:
Creates the UnstableButton.
- variants: primary (default), stroked, ghost, and destructive.
- sizes: small, medium (default), and  large

Expanded unstableFontFaces to have `medium` font weight for inter regular and italic

Part of #350


WilliamKelley:
Opened #376 and removed TODO in PR description.

shadoath:
> This linear-gradients seem unnecessary, it's not even a gradient (same colors) so its just an overlay? Maybe just color-dropper the resulting color, and use that instead... Idk

Here is what I'm thinking, in Figma it is defined like this for the hover color
Screen Shot 2022-04-05 at 14 02 18

Looks like 
```
import { lighten } from '@material-ui/core';
lighten(theme.unstable_palette.brand.blue, 0.1)
```
does the trick! 

It takes goes from `#0A4872` to `#225A80` or `rgb(34, 90, 128)`

WilliamKelley:
> Looks like
> ```
> import { lighten } from '@material-ui/core';
> lighten(theme.unstable_palette.brand.blue, 0.1)
> ```
> does the trick!

Nice, that works. We'll have to be careful with the import, use 
```ts
import { lighten } from '@material-ui/core/styles`
```
instead so Meteor bundling doesn't include all of the pkg by default for the one util.

shadoath:
@WilliamKelley Ready for you to take another look. I'm struggling to figure out how to get the icons styling to work it does not seem to be targeting correctly.

WilliamKelley:
> @WilliamKelley Ready for you to take another look. I'm struggling to figure out how to get the icons styling to work it does not seem to be targeting correctly.

Sure, will take a look

libs/spark/src/Unstable_Button/Unstable_Button.stories.tsx
```diff
@@ -0,0 +1,245 @@
...
+  extends Omit<
+    Unstable_ButtonProps,
+    | 'disableElevation'
+    | 'disableFocusRipple'
+    | 'centerRipple'
+    | 'disableRipple'
+    | 'disableTouchRipple'
+    | 'focusRipple'
+    | 'tabIndex'
+    | 'TouchRippleProps'
```
WilliamKelley, line 18: You can omit these in the actual props. We don't want them to appear for consumers either.

libs/spark/src/Unstable_Button/Unstable_Button.stories.tsx
```diff
@@ -0,0 +1,245 @@
...
+interface Sb_Unstable_ButtonProps
+  extends Omit<
+    Unstable_ButtonProps,
+    | 'disableElevation'
+    | 'disableFocusRipple'
+    | 'centerRipple'
+    | 'disableRipple'
+    | 'disableTouchRipple'
+    | 'focusRipple'
+    | 'tabIndex'
```
WilliamKelley, line 17: Don't omit this! this is necessary

libs/spark/src/Unstable_Button/Unstable_Button.stories.tsx
```diff
@@ -0,0 +1,245 @@
+import * as React from 'react';
+import type { Meta, Story } from '@storybook/react/types-6-0';
+import { ChevronDown, AlertCircle } from '@prenda/spark-icons';
```
WilliamKelley, line 3: use "Unstable_" icons, or copy the `unstable_createSvgIcon` invocation call of another story.

libs/spark/src/Unstable_Button/Unstable_Button.tsx
```diff
@@ -0,0 +1,244 @@
...
+import { buildVariant } from '../theme/typography';
+import { lighten, darken } from '@material-ui/core/styles';
+
+export interface Unstable_ButtonTypeMap<
+  // eslint-disable-next-line @typescript-eslint/ban-types
+  P = {},
+  D extends React.ElementType = 'button'
+> {
+  props: P &
+    Omit<MuiButtonProps, 'classes' | 'variant'> & {
```
WilliamKelley, line 18: Here's where you'd omit all the stuff

libs/spark/src/Unstable_Button/Unstable_Button.tsx
```diff
@@ -0,0 +1,244 @@
...
+          backgroundColor: theme.unstable_palette.neutral[600],
+        },
+        // TODO Loading
+        '&:disabled': {
+          backgroundColor: theme.unstable_palette.neutral[80],
+          color: theme.unstable_palette.neutral[100],
+        },
+        '&.Mui-focusVisible, &:focus-visible': {
+          boxShadow: `0px 0px 2px 4px ${theme.unstable_palette.teal[300]}`,
+        },
```
WilliamKelley, line 87: This is common to all variants so you can extract to unconditionally apply

libs/spark/src/Unstable_Button/Unstable_Button.tsx
```diff
@@ -0,0 +1,244 @@
...
+  '"Inter", sans-serif',
+  "'cv05' 1, 'ss03' 1"
+);
+const useStyles = makeStyles<Unstable_ButtonClassKey>(
+  (theme) => ({
+    root: (props: Unstable_ButtonProps) => ({
+      borderRadius: 4,
+      ...(props.variant === 'primary' && {
+        backgroundColor: theme.unstable_palette.brand.blue,
+        color: theme.palette.common.white,
```
WilliamKelley, line 70: Text color and font size are specific to the dom tag with the `label` class key, not `root`.

libs/spark/src/Unstable_Button/Unstable_Button.tsx
```diff
@@ -0,0 +1,244 @@
...
+          backgroundColor: lighten(theme.unstable_palette.brand.blue, 0.1),
+        },
+        '&:active': {
+          backgroundColor: darken(theme.unstable_palette.brand.blue, 0.2),
+        },
+        '&[aria-expanded="true"]': {
+          backgroundColor: theme.unstable_palette.neutral[600],
+        },
+        // TODO Loading
+        '&:disabled': {
```
WilliamKelley, line 81: condition on the prop here, not pseudo-class

libs/spark/src/Unstable_Button/Unstable_Button.tsx
```diff
@@ -0,0 +1,244 @@
...
+        '&:hover': {
+          backgroundColor: lighten(theme.unstable_palette.brand.blue, 0.1),
+        },
+        '&:active': {
+          backgroundColor: darken(theme.unstable_palette.brand.blue, 0.2),
+        },
+        '&[aria-expanded="true"]': {
+          backgroundColor: theme.unstable_palette.neutral[600],
+        },
+        // TODO Loading
```
WilliamKelley, line 80:
```suggestion
```

Unnecessary TODO, the LoadingButton will be a separate component, not a conditional style

libs/spark/src/Unstable_Button/Unstable_Button.tsx
```diff
@@ -0,0 +1,244 @@
...
+      variant = 'primary',
+      size = 'medium',
+      ...other
+    } = props;
+
+    const classes = useStyles({ variant, size });
+
+    return (
+      <MuiButton
+        classes={{ root: clsx(classes.root, classesProp?.root) }}
```
WilliamKelley, line 232: you need to specify the other class keys here, not just `root` -- so `startIcon`, `endIcon`, `label` all get the same treatment.

WilliamKelley (changes requested):
The medium font files are named wrong, need to prepend "inter-" -- can see the error in your console.

To debug this, you can use the DevTools to go thru the styles applied to the startIcon wrapper. You would see that there's no class names including "MuiSparkUnstable_Button", which means they're not getting passed on.

shadoath:
@WilliamKelley Bring on the next round of reviews!

libs/spark/src/Unstable_Button/Unstable_Button.tsx
```diff
@@ -0,0 +1,277 @@
...
+      // -2px accounts for label vs icon size difference
+      // have to specify individually to override defaults
+      marginTop: '-2px',
+      marginBottom: '-2px',
+      marginLeft: '8px',
+      marginRight: 0,
+      color: 'inherit',
+      ...(props.size === 'small' && {
+        fontSize: theme.typography.pxToRem(16),
+        lineHeight: theme.typography.pxToRem(16),
```
WilliamKelley, line 232:
```suggestion
      color: 'inherit',
      lineHeight: 1,
      ...(props.size === 'small' && {
        fontSize: theme.typography.pxToRem(16),
```
Line height's should be unitless. They become a coefficient of font size.

libs/spark/src/Unstable_Button/Unstable_Button.tsx
```diff
@@ -0,0 +1,277 @@
...
+      }),
+    }),
+    endIcon: (props: Unstable_ButtonProps) => ({
+      display: 'inherit',
+      // -2px accounts for label vs icon size difference
+      // have to specify individually to override defaults
+      marginTop: '-2px',
+      marginBottom: '-2px',
+      marginLeft: '8px',
+      marginRight: 0,
```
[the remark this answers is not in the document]
WilliamKelley: Okay on revisit, I think I know why you specified this. Is it because the text baseline in Figma is different than the Web so the text seems like it's "too high" in comparison?

If so, here's the workaround I found:

```
{
  label: { marginTop: 2, marginBottom: -2 }, // shift text baseline down
  startIcon: { marginTop: -2 }, // shift icon baseline up
  endIcon: { marginTop: -2 }
}
```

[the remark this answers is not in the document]
WilliamKelley: Some references: 
- Figma with an arbitrary box to show the extension of the "baseline" and the line of "cap-height"
- Implementation before margin adjustments
- and after
image
image
image

WilliamKelley (changes requested):

libs/spark/src/Unstable_Button/Unstable_Button.tsx
```diff
@@ -0,0 +1,277 @@
...
+        },
+        '&$disabled': {
+          backgroundColor: theme.unstable_palette.neutral[80],
+        },
+        '&.Mui-focusVisible, &:focus-visible': {
+          backgroundColor: theme.palette.common.white,
+        },
+      }),
+      ...(props.variant === 'destructive' && {
+        border: '2px solid transparent',
```
WilliamKelley, line 137:
```suggestion
```
Not necessary?

WilliamKelley (changes requested):

libs/spark/src/Unstable_Button/Unstable_Button.tsx
```diff
@@ -0,0 +1,277 @@
...
+        },
+        '&[aria-expanded="true"]': {
+          backgroundColor: theme.unstable_palette.neutral[600],
+        },
+        '&$disabled': {
+          backgroundColor: theme.unstable_palette.neutral[80],
+        },
+        '&.Mui-focusVisible, &:focus-visible': {
+          backgroundColor: theme.palette.common.white,
+        },
```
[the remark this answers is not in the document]
shadoath: Yup saw that when I got in today and nuked it.

libs/spark/src/Unstable_Button/Unstable_Button.tsx
shadoath: Correct

libs/spark/src/Unstable_Button/Unstable_Button.tsx
```diff
@@ -0,0 +1,277 @@
...
+        '&[aria-expanded="true"]': {
+          color: theme.palette.common.white,
+        },
+      }),
+      ...(props.variant === 'ghost' && {
+        color: theme.unstable_palette.brand.blue,
+        '&[aria-expanded="true"]': {
+          color: theme.palette.common.white,
+        },
+        '&$disabled': {
```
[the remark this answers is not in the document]
shadoath: I got this to work with `':disabled > & ': {` should I go to using the `...(props.disabled && {` way instead?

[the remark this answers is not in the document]
WilliamKelley: Yes

libs/spark/src/Unstable_Button/Unstable_Button.tsx
```diff
@@ -0,0 +1,277 @@
...
+        '&[aria-expanded="true"]': {
+          backgroundColor: theme.unstable_palette.neutral[600],
+        },
+        '&$disabled': {
+          backgroundColor: theme.unstable_palette.neutral[80],
+          color: theme.unstable_palette.neutral[100],
+        },
+      }),
+      ...(props.size === 'small' && {
+        ...buttonFontVariantSmall,
```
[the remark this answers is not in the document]
WilliamKelley: Only the typography properties, the `padding` still applies to the `root`. So
```
{
  root: {
    ...(props.size === 'small' && { padding: '...' })
  },
  label: {
    ...(props.size === 'small' && buttonFontVariantSmall)
  }
}
```

libs/spark/src/Unstable_Button/Unstable_Button.tsx
```diff
@@ -0,0 +1,270 @@
...
+          endIcon: clsx(classes.endIcon, classesProp?.endIcon),
+          label: clsx(classes.label, classesProp?.label),
+        }}
+        disableFocusRipple
+        disableRipple
+        disableTouchRipple
+        focusRipple={false}
+        ref={ref}
+        {...other}
+      />
```
WilliamKelley, line 265:
```suggestion
    const {
      classes: classesProp,
      disabled,
      variant = 'primary',
      size = 'medium',
      ...other
    } = props;

    const classes = useStyles({ disabled, variant, size });

    return (
      <MuiButton
        classes={{
          root: clsx(classes.root, classesProp?.root),
          startIcon: clsx(classes.startIcon, classesProp?.startIcon),
          endIcon: clsx(classes.endIcon, classesProp?.endIcon),
          label: clsx(classes.label, classesProp?.label),
        }}
        disabled={disabled}
        disableFocusRipple
        disableRipple
        disableTouchRipple
        focusRipple={false}
        ref={ref}
        {...other}
      />
```

WilliamKelley: styles depend on disabled now, so destructure and stuff

libs/spark/src/Unstable_Button/Unstable_Button.tsx
```diff
@@ -0,0 +1,279 @@
...
+          backgroundColor: theme.unstable_palette.red[600],
+        },
+        '&:active': {
+          backgroundColor: darken(theme.unstable_palette.red[700], 0.2),
+        },
+        '&[aria-expanded="true"]': {
+          backgroundColor: theme.unstable_palette.neutral[600],
+          '& span': {
+            color: theme.palette.common.white,
+          },
```
WilliamKelley, line 148: move this under label with `[aria-expanded="true'] > &` like the others?

WilliamKelley:
Pushed a small change to reduce the text baseline shift by a pixel when the size is small -- it was too far down in that scenario.

WilliamKelley:
@shadoath Everything that's up looks good. Great job.

Now for some fairly major changes in naming I'm just realizing. Like in `Unstable_Input` and related components, we have the opportunity to rename the `startIcon` and `endIcon` props to `leadingIcon` and `trailingIcon`, which we should do. We can either translate between the two or re-implement the functionality. i.e. 
```jsx
// translate
<Unstable_Button startIcon={leadingIcon}>
  {children}
</Unstable_Button>
// re-implement
<Unstable_Button>
  {leadingIcon}
  {children}
  {trailingIcon}
</Unstable_Button>
```

This re-implementation will have ripple effects as the icons will no longer be children of the "label" `span` tag. That's only in MUI v4 because of an outdated restriction of `button` tags not being able to be flexboxes -- that's not the case anymore, we can set `display: flex` on the root (if it's not already). This will simplify our text baseline shift (we can just shift the label I think) and we won't have to override default mui styles on `startIcon` and `endIcon` because those slots aren't being applied by the underlying component!

We'll also have to have a `leadingAvatar` prop since that has it's own styling that's not that same as a `leadingIcon`.

Note that for each, the props signature and class keys will change.

We can decide to do this in this PR, or leave it for a next PR, but it should be done.

shadoath:
@WilliamKelley thank you, I do see how we could transition to the leadingIcon and trailingIcon but would prefer that be in another PR.

WilliamKelley:
@shadoath Okay, will convert to ready and confirm final changes in Chromatic

WilliamKelley (approved):

slingner (approved):
Nice!!

WilliamKelley (approved):

WilliamKelley:
Missed the changelog in this PR, @shadoath will follow up in separate PR

PR #378, opened 2022-03-31
refactor(unstable_fontFaces): only .woff2 files; versioned asset paths

### LINKED ISSUE #376: Unnecessary font files
Remove `.woff` font files in `libs/spark/public`. Also remove their references in `libs/spark/theme/(fontFaces|unstable_fontFaces).ts`

Only IE doesn't support `.woff2` ([see MDN](https://developer.mozilla.org/en-US/docs/Web/Guide/WOFF#browser_compatibility)), and, say it with me, we don't care about IE!

### LINKED ISSUE #377: Browsers cache old font files
Also, need to flatten fonts filenames to mitigate browser caching old versions by domain and file name. i.e. fonts/inter-bold-v1.woff2 instead of fonts/inter/bold.woff2

WilliamKelley:
Close #376 
Close #377


libs/spark/src/theme/unstable_fontFaces.ts
```diff
@@ -1,145 +1,123 @@
 import * as CSS from 'csstype';
 
+/**
+ * A small, internal utility to generate a standard `src` property value for `@font-face` declarations.
+ *
+ * MAINTAINER NOTE: Whenever an expected asset file changes (in this case a font file), the version number must be incremented. Otherwise, browsers may use the previously cached file for a long time.
+ */
+const buildSrc = (filename: string) =>
+  `url(/pds-assets-v1/fonts/${filename}.woff2) format("woff2")`;
+
```
shadoath, line 10: Good addition. Worth making a note in the Readme I think.

shadoath (approved):

slingner (approved):

libs/spark/src/theme/unstable_fontFaces.ts
WilliamKelley: Yeah it's on my list for the README, for right now, heads-up in the changelog for affected versions is fine.

PR #380, opened 2022-03-31
feat!: separate FontFacesBaseline from CssBaseline; add unstable versions

WilliamKelley:
Separated the injection of Mui's "css reset" from our injection of "font-face" decalarations. I also stopped wrapping Mui's component. It's dead simple, and avoids us having to overwrite their values. 

FontFacesBaseline now gets the values directly, whereas before it would have gotten them through the theme. I realized that indirection was unnecessary. 

Both are not included in `SparkThemeProvider` anymore, so consumers are advised to render them separately. While it's more to do, I think it's much better for the consumer to separate the idea of these baselines as one-time global styles, rather than a provider which contains a dynamic value.

---

I should mention that this is part of a larger effort to remove `SparkThemeProvider` as a global storybook decorator. All these baselines **should** be global decorators, but not the `ThemeProvider` / `StylesProvider`. Those should be rendered only if the given component needs it. While mostly all of our stable components need it, a core objective of our unstable components is to **not** need it. We can make sure they don't error through tests, but ensuring correct styling without `ThemeProvider` will be done through Storybook, hence this change.


shadoath (approved):

slingner (approved):
Nice!

PR #382, opened 2022-03-31
chore: remove SparkThemeProvider as global decorator

WilliamKelley:
Blocked by #380, #381 

Will help us develop PDS v2 / `Unstable_*` components. If an underlying component relies on the overrides/default props through the `theme`, then the `ThemeProvider` acts behind-the-scenes. This avoids that and _ensures correct styling_ with `SparkThemeProvider` being an ancestory, like stated in the CHANGELOG. 

Of course, we need to ALSO make sure it renders the same WITH `SparkThemeProvider` being an ancestor, so I'm planning to make a single "(ThemeProvider)" story for each v2 component and do some manual QA during dev.

### DIFF
--- a/libs/spark/.storybook/preview.js
+++ b/libs/spark/.storybook/preview.js
@@ -3,7 +3,6 @@ import { addDecorator } from '@storybook/react';
 import {
   CssBaseline,
   FontFacesBaseline,
-  SparkThemeProvider,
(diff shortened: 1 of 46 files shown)


shadoath (approved):
Feels like Christmas with all these decorations going in.

slingner (approved):

PR #396, opened 2022-04-26
feat(Unstable_Tag): initial implementation

WilliamKelley:
Part of #350 

I winged some things that the Figma did not specify like styling for the "active" pseudo state, the "clickable" styling, plain "icon" prop, etc.

I had to keep the "cross-product" layout for the stories because the complex underlying styles mean we should validate all possibilities for conflicts, and creating individual stories for each is not only way too much code, unfortunately, it's also 2 x 2 x 7 = 28 times more stories for _each_ current "cross-product" story (which there are 10 of). (The special Unicode char in the story names is the a cross product symbol.)

### DIFF
--- a/libs/spark/src/Unstable_Tag/Unstable_Tag.stories.tsx
+++ b/libs/spark/src/Unstable_Tag/Unstable_Tag.stories.tsx
@@ -1,0 +1,251 @@
+import * as React from 'react';
+import type { Meta, Story } from '@storybook/react/types-6-0';
+import { Unstable_Tag, unstable_createSvgIcon } from '..';
+import { sparkThemeProvider } from '../../stories';
+import { Unstable_TagProps } from './Unstable_Tag';
+
+export const SbUnstable_Tag = Unstable_Tag;
+
+const Filter = unstable_createSvgIcon(
+  <path d="M2.25 6A.75.75 0 0 1 3 5.25h18a.75.75 0 0 1 0 1.5H3A.75.75 0 0 1 2.25 6Zm3 6a.75.75 0 0 1 .75-.75h12a.75.75 0 0 1 0 1.5H6a.75.75 0 0 1-.75-.75ZM10 17.25a.75.75 0 0 0 0 1.5h4a.75.75 0 0 0 0-1.5h-4Z" />,
+  'Sb_Filter'
+);
+
+const emptyFn = () => {
+  return;
+};
+
+export default {
+  title: '@ps/Unstable_Tag',
+  component: SbUnstable_Tag,
+  excludeStories: ['SbUnstable_Tag'],
+  parameters: {
+    actions: {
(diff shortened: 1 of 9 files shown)


libs/spark/src/Unstable_Tag/Unstable_Tag.tsx
```diff
@@ -0,0 +1,285 @@
...
+        fontSize: theme.typography.pxToRem(16),
+      }),
+      ...(props.size === 'small' && {
+        fontSize: theme.typography.pxToRem(12),
+      }),
+    }),
+    avatar: {
+      // blocked until `Unstable_Avatar` initial implementation
+    },
+  }),
```
shadoath, line 242: Wow, that's a lot of styling.

libs/spark/src/Unstable_Tag/Unstable_Tag.stories.tsx
```diff
@@ -0,0 +1,251 @@
...
+      // override default actions regex
+      //  deleteIcon rendering is based on onDelete being supplied, so substitute
+      //  custom argType control below.
+      argTypesRegex: '^on(?!Delete).*',
+    },
+  },
+  argTypes: {
+    icon: {
+      control: 'select',
+      options: ['undefined', '$', 'Filter'],
```
shadoath, line 33: What is the '$' option here? it didn't do anything in storybook when I clicked it.

libs/spark/src/Unstable_Tag/Unstable_Tag.tsx
```diff
@@ -0,0 +1,285 @@
...
+
+const Unstable_Tag: OverridableComponent<Unstable_TagTypeMap> = React.forwardRef(
+  function Unstable_Tag(props, ref) {
+    const {
+      classes: classesProp,
+      color = 'neutral',
+      onClick,
+      onDelete,
+      size = 'medium',
+      variant = 'bold',
```
shadoath, line 254: Should the controls default to these values as well?

libs/spark/src/Unstable_Tag/Unstable_Tag.stories.tsx
WilliamKelley: Wasn't meant to be there, i copied the object from another story and forgot to remove that.

slingner (approved):

libs/spark/src/Unstable_Tag/Unstable_Tag.tsx
WilliamKelley: I don't want them to be. It'd be nice for the documentation, but the point of our "default" story is to validate the styling of when no props are specified. and having the controls explicitly default to that would cause them to not default, but instead take on that value explicitly.

shadoath (approved):

PR #397, opened 2022-04-27
feat(Unstable_Select): render tags when multiple

WilliamKelley:
Part of #350 

When `multiple` is true, `Unstable_Select` will now render the value (values) as tags (`Unstable_Tag`). This prevents a normal, comma-separated text representation, but I believe that's good -- this is the approach that MUI takes with `Autocomplete`. The user can always override it.

Adds the `getTagProps` prop for any customizations a user may want to make that we can't anticipate, like displaying custom labels. 

Unfortunately, we can't make these tags deletable ourselves because the `onChange` handler is typed for DOM Events, not a custom value. I don't think we want to get into the business of artificially creating dom events / objects or adding a parallel change handler like Autocomplete's. Besides, the user can implement that functionality through `getTagProps`.

Since the tags are taller than normal text, some padding manipulation has to happen in the styles unfortunately. Also, you'll notice a very special html char reference when there's no tags but multiple is set. That's because the height and layout will shift unless the span's vertical alignment takes effect, which only happens when there's something inside of it! The character is a "no width white space" char. Setting it through `dangerouslySetInnerHtml` makes it not show up in the DOM i think? After being baffled about what was missing (because it seems like an empty span in the DevTools DOM), I finally dug deep enough into MUI to see them define this behavior.


WilliamKelley:
> When not using the MultiSelect the LeadingEl and the text overlap.

Oops, will take a look.

PR #398, opened 2022-04-27
fix(Unstable_Button): override previous theme styles

shadoath:
- Specifically targeting when the theme Provider of spark v1 has styling overrides.

### DIFF
--- a/libs/spark/CHANGELOG.md
+++ b/libs/spark/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## [vNext](https://github.com/prenda-school/prenda-spark/compare/v1.0.0-alpha.5...vNext) (yyyy-mm-dd)
 
+- **Unstable_Button**
+  - Update styling on Unstable button to match Figma mockups. Required overriding the SparkThemeProvider styles.
+
 ## [v1.0.0-alpha.5](https://github.com/prenda-school/prenda-spark/compare/v1.0.0-alpha.4...v1.0.0-alpha.5) (2022-04-25)
 
 ### Fixes
--- a/libs/spark/CHANGELOG.md
+++ b/libs/spark/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 - **Unstable_Button**
   - Override conflicting styles from previous `theme` (e.g. when rendered under `SparkThemeProvider`) for pseudo states like hover, disabled, etc.
+
 ## [v1.0.0-alpha.5](https://github.com/prenda-school/prenda-spark/compare/v1.0.0-alpha.4...v1.0.0-alpha.5) (2022-04-25)
 
 ### Fixes
--- a/libs/spark/CHANGELOG.md
+++ b/libs/spark/CHANGELOG.md
@@ -2,9 +2,10 @@
 
 ## [vNext](https://github.com/prenda-school/prenda-spark/compare/v1.0.0-alpha.5...vNext) (yyyy-mm-dd)
 
-- **Unstable_Button**
-  - Update styling on Unstable button to match Figma mockups. Required overriding the SparkThemeProvider styles.
+### Fixes
 
+- **Unstable_Button**
+  - Override conflicting styles from previous `theme` (e.g. when rendered under `SparkThemeProvider`) for pseudo states like hover, disabled, etc.
 ## [v1.0.0-alpha.5](https://github.com/prenda-school/prenda-spark/compare/v1.0.0-alpha.4...v1.0.0-alpha.5) (2022-04-25)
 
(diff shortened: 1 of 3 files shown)


libs/spark/CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## [vNext](https://github.com/prenda-school/prenda-spark/compare/v1.0.0-alpha.5...vNext) (yyyy-mm-dd)
 
+- **Unstable_Button**
+  - Update styling on Unstable button to match Figma mockups. Required overriding the SparkThemeProvider styles.
+
```
WilliamKelley, line 7:
```suggestion
### Fixes

- **Unstable_Button**
  - Override conflicting styles from previous `theme` (e.g. when rendered under `SparkThemeProvider`) for pseudo states like hover, disabled, etc.
```

libs/spark/src/Unstable_Button/Unstable_Button.tsx
```diff
@@ -79,13 +79,20 @@ const buttonFontVariantLarge = buildVariant(
...
       borderRadius: 4,
       borderStyle: 'solid',
       borderWidth: 1,
       '&.Mui-focusVisible, &:focus-visible': {
         boxShadow: `0px 0px 2px 4px ${theme.unstable_palette.teal[300]}`,
       },
+      // Double specificity to override V1 theme style
+      '&&:disabled': {
+        opacity: 'unset',
+      },
```
WilliamKelley, line 95:
```suggestion
      // override V1 theme style
      '&&:disabled': {
        opacity: 'unset',
      },
```
It's only to override the style. Not an existing style that needs double specificity.

libs/spark/src/Unstable_Button/Unstable_Button.tsx
```diff
@@ -102,11 +109,13 @@ const useStyles = makeStyles<Unstable_ButtonClassKey>(
         },
       }),
       ...(props.variant === 'stroked' && {
-        borderColor: `${theme.unstable_palette.neutral[90]}`,
-        backgroundColor: 'transparent', // theme.palette.common.white,
-        // '&:hover': {
-        //   backgroundColor: theme.unstable_palette.neutral[70],
-        // },
+        '&&': {
```
WilliamKelley, line 112:
```suggestion
        // double specificity to override v1 theme style
        '&&': {
```

libs/spark/src/Unstable_Button/Unstable_Button.tsx
```diff
@@ -195,6 +204,9 @@ const useStyles = makeStyles<Unstable_ButtonClassKey>(
         ...(props.disabled && {
           color: theme.unstable_palette.neutral[100],
         }),
+        '&:hover > & ': {
+          color: theme.palette.common.white,
+        },
```
WilliamKelley, line 209:
```suggestion
```
Pretty sure this was unnecessary.

WilliamKelley (changes requested):
Title changed to comply to conventional commits.

libs/spark/src/Unstable_Button/Unstable_Button.tsx
shadoath: Correct...whoops

shadoath:
@WilliamKelley, I had forgot about that. It is just so nice to hit merge commit. 
I could turn the PR to draft and do it though then bring it back out to trigger only one chromatic merge right?

shadoath:
Or use the 'batch' ill use that more.

WilliamKelley:
Oh, it looks like even successful runs have that error message, but it doesnt exit the process with an error code and fail the job, it exits fine.

So we should follow with the workaround described in the issue

shadoath:
So we need to do the following: 
```
variables:
  GIT_CHECKOUT: 'false'

before_script:
  - git checkout "$CI_COMMIT_REF_NAME"
  ```

WilliamKelley:
> So we need to do the following?
> 
> ```
> variables:
>   GIT_CHECKOUT: 'false'
> 
> before_script:
>   - git checkout "$CI_COMMIT_REF_NAME"
> ```

I believe those are GitLab specific, so it would be the equivalent GitHub Actions env var? I'm not sure. 

Regardless though, it seems like it succeeded once you added an empty line between a list item and a heading! Which makes sense because the absence of that would cause format:check to fail. So it just has a poor output.

WilliamKelley (approved):

WilliamKelley:
@slingner review bump. we can get this released today

slingner (approved):
🚀

PR #399, opened 2022-04-28
Alert export

shadoath:
Add alert to the main export file

**Issue** now `export type Color = 'success' | 'info' | 'warning' | 'error';` is exported twice from `Alert` and `Unstable_SectionMessage`

### DIFF
--- a/libs/spark/src/index.ts
+++ b/libs/spark/src/index.ts
@@ -1,3 +1,6 @@
+export { default as Alert } from './Alert';
+export * from './Alert';
+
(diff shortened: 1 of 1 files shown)


shadoath:
@WilliamKelley how do you want to handle the double export? 
- Move `Color` to its own file
- Make them unique for each file it is in?
- Something else?

WilliamKelley:
Why export Alert? PDS doesnt specify a generalAlert component, and Banner / Section Message are already implemented.

shadoath:
Roger, will use the Banner/ Section message instead.

PR #401, opened 2022-05-08
feat: add example app with nextjs

WilliamKelley:
Added an "example" app for Next.js. This was a playground for debugging recent issues like the Marketplace had with JSS dynamic styles and React strict mode.


shadoath (approved):
Nice that the work you did to get this working can be added here as a good future example.

slingner (approved):

Issue #403, opened 2022-05-12
Unstable_SectionMessage onClose prop does not work


shadoath:
When using a section message we have the onClose method defined but it is not wired up to anything.

WilliamKelley:
This was intentional since the `Unstable_IconButton` is not developed yet.

WilliamKelley:
Fixed by #404

PR #404, opened 2022-05-12
Add onClick handler to section message icon for closing

shadoath:
Noticed we didn't get a close when hitting the `x` in the Portal.

Tracked it down to no action being taken for the `onClose` function. 

Also added a console.log("closing") to the story, but don't like that. I think there is a better solution.

Fixing #403


WilliamKelley:
This was because it needs to be an instance of a button, i.e. `Unstable_IconButton`, variant `"ghost"`.

An `onClick` on a div is not accessible.

shadoath:
@WilliamKelley thoughts on what to do next here? 
Do we do a small fix like this? or wait till we have the Unstable_IconButton?

WilliamKelley:
> @WilliamKelley thoughts on what to do next here? Do we do a small fix like this? or wait till we have the Unstable_IconButton?

We can move forward with this temporary stand-in (note: it's not a "fix"; it was never broken, rather it was intentionally left unimplemented).

It'd be nicer to take the hour to do `Unstable_IconButton` and skip this -- I would think this will be quick given most of the work can be copied from `Unstable_Button`. But I understand either way.

libs/spark/src/Unstable_SectionMessage/Unstable_SectionMessage.stories.tsx
```diff
@@ -23,6 +23,7 @@ SeverityInfo.args = { severity: 'info' };
 SeverityInfo.storyName = 'severity=info';
 
 const onClose = () => {
+  console.log('CLOSING');
```
WilliamKelley, line 26:
```suggestion
```

WilliamKelley (approved):

shadoath:
@slingner  bump. William has implemented the unstable Button

slingner (approved):

PR #405, opened 2022-05-16
fix(unstable_typography): wrong weight for T22 variant

WilliamKelley:
Primarily fixes the "T22" typography variant font weight. 

Also fixes the fact that the component wasn't being typed in the right way and the resulting props were losing information, like `variant: string` instead of the true `variant: "display" | "T32" | ...`.

Also rewrites the stories for `Unstable_Typography`. The big example section is already reflected in the story for the theme property story, i.e. `theme/unstable_typography`. The component story now adheres to single stories that just go thru the prop values.


WilliamKelley:
@slingner @shadoath review bump

slingner (approved):

shadoath (approved):

PR #406, opened 2022-05-20
chore(Unstable_Button): standardize stories

WilliamKelley:
Refactors the stories file for Unstable_Button:

- Custom names for all stories that follow the lowercase prop combo format
- Custom grid template replaced with simple flexboxes (like Unstable_Tag)
- Removed combos that weren't useful (like start/end icon with pseudo states other than disabled -- there's no style changes expected, and no good reason to be concerned).

Each story is still duplicated with a decorating with SparkThemeProvider -- from this, the outstanding issues are clear:

- start/end icon margin shift back-fires under STP
- disabled is wrong under STP

This PR will not aim to fix this issues as the Chromatic snapshots will be reset by these changes.


slingner (approved):

shadoath (approved):
Nice clean up on this!

PR #408, opened 2022-05-24
feat(Unstable_IconButton): initial implementation

WilliamKelley:
Part of #350 

Added a new decorator in an effort to have Chromatic fully capture the focus indicator -- [margin didn't work in the previous attempt for "Unstable_Button"](https://www.chromatic.com/component?appId=60b7f55f55fbd4004993da4c&csfId=ps-unstable-button--size-by-variant-focus-visible&buildNumber=722&k=6287d0d775c890004a1ce17e-1200-snapshot-true&h=15&b=-5); I'm assuming because either their tool disregards it or there's some margin collapsing going on or something.

_Edit: [the new decorator works](https://www.chromatic.com/test?appId=60b7f55f55fbd4004993da4c&id=628d4ef1de6ce4003a0a8b64)!_

PR #409, opened 2022-05-25
chore(Unstable_Button): fix snapshots for focus indicator

WilliamKelley:
Implements the new decorator from #408 to ensure the focus indicator is fully contained in Chromatic snapshots.


libs/spark/src/Unstable_Button/Unstable_Button.stories.tsx
```diff
@@ -65,24 +66,21 @@ const SizeByVariantTemplate = (args) => (
...
         {variants.map((variant) => (
           <Unstable_Button
             key={variant}
             {...args}
             variant={variant}
             size={size}
-          >
-            Label
-          </Unstable_Button>
+          />
```
shadoath, line 83: Args might include 'children' that's why you removed the `Label` child below?

shadoath (approved):

slingner (approved):

PR #413, opened 2022-06-01
feat(Unstable_Button): improve text baseline shift

WilliamKelley:
The alignments were still off from the attempt to shift the text baseline. Instead of messing with the entire label to adjust the text (requires re-adjusting the icon), we just add a wrap the text in an extra tag and shift that instead. This is better because the icon is always aligned correctly in the button, it's just the text that has a higher baseline than normal (and doesn't appear vertically centered without the shift).

**Before**
image

**After**
image


shadoath (approved):
I like this approach over the pixel changes at so many levels.

PR #414, opened 2022-06-01
refactor: extract Unstable_Alert from Unstable_SectionMessage

WilliamKelley:
In prep for `Unstable_Banner`, which is another alert like `Unstable_SectionMessage` with different placement/styling, I've pulled out an unstyled, common skeleton that they can share: `Unstable_Alert`.

Also makes a couple changes to the styles of `Unstable_SectionMessage`, mainly setting `display: flex` on the severity icon slot in order to remove the automatic spacing for text ascender and descenders which is not needed since it's an icon. This is an obscure part of CSS! (the effect was that even though the icon height was 24px, the box layout height would be 27-ish px -- this would cause ripple effects across the component styles.


slingner (approved):
Very nice! I thought there was still a small alignment issue with Section Message but couldn't place where it was coming from. Thanks for the update and explanation.

PR #419, opened 2022-06-04
feat(Unstable_Avatar): initial implementation

WilliamKelley:
Initial implementation of PDS v2 Avatar.

I'll give an illustration of the height/width a11y comment I left. Here's a compilation of screenshots

- top row: browser font size is 16px (responsive-ness is moot)
- middle row: browser font size is 56px and the height/width is not responsive (i.e. `height: 64, width: 64`)
- bottom row: browser font size is 56px and the height/width is responsive (i.e. `minHeight: 64, height: '...rem', ...`)

image


shadoath (approved):
Very nice!  I like what you showed us with the Ally work on this too.

WilliamKelley:
Closes #446

PR #421, opened 2022-06-07
feat(Unstable_Button)!: replace (start|end)Icon with (leading|trailing)Icon

WilliamKelley:
(prerequisite work to supporting `leadingAvatar`)

Replaces use of Mui's underlying prop with our custom named prop and slots. Allows us to get rid of validating styles with leading/trailing icon since it's no longer subject to underlying Mui and v1 theme behavior.

Also
- simpler re-typing of the component so Storybook can pick up the props typing
- start moving icons used in stories to central file like decorators.


slingner (approved):

WilliamKelley:
@shadoath bump, mighta missed this one

libs/spark/src/Unstable_Button/Unstable_Button.stories.tsx
```diff
@@ -1,45 +1,41 @@
...
+} from '../../stories';
 
-// eslint-disable-next-line @typescript-eslint/no-empty-interface
-interface Sb_Unstable_ButtonProps
-  extends Omit<Unstable_ButtonProps, 'tabIndex'> {}
-
-export const Sb_Unstable_Button = (props: Sb_Unstable_ButtonProps) => (
-  <Unstable_Button {...props} />
-);
+export const _retyped = Unstable_Button as typeof Unstable_Button;
```
shadoath, line 11: Why do this?

shadoath (approved):

libs/spark/src/Unstable_Button/Unstable_Button.stories.tsx
WilliamKelley: the auto-controls storybook plugin has trouble picking up prop types of default exports. so sometimes you have to re-export as a named export and use that instead, but also re-apply the props/typing. this is the shortest way ive found to do so.

PR #434, opened 2022-06-11
feat: improve component typings

WilliamKelley:
Two things:

- props that are assigned types _inside_ of the `forwardRef` function will not have the best typing -- things like discriminated unions of constant strings will show up as just `string`
- strictly typing a ref is problematic when the component is overridable -- the ref should come from the component, but the types can't be that smart. not all these components are overridable, but the principal is sound.

So, all refs are typed as unknown, and ref and props typings are given via generics to `forwardRef`.


shadoath (approved):
I think you need to go over this with me in person as to why we need this.

WilliamKelley:
> I think you need to go over this with me in person as to why we need this.

Sure. Also if it helps, the second paragraph of #406 is an example of the problem.

slingner (approved):

PR #435, opened 2022-06-11
chore: standardize styles

WilliamKelley:
- Add block comment groupings to prop conditional spreads
- Pull focus styles before all props conditions because it's always constant
- Push disabled styles to be **last** so they override anything before it.
- Extract any common style blocks among conditional spreads
- Use `text.disabled` instead of `neutral[0]` where appropriate for text color
- Use `background.inverse` instead of `neutral[600]` where appropriate for background color
- Condition on `aria-expanded` as a prop instead of a DOM attribute


slingner (approved):

shadoath (approved):
I scrolled to the bottom and my eyes saw most of the code as it went by.
And I have a high level of trust in your PRs.

PR #437, opened 2022-06-11
feat: add error prop/styles for radio, checkbox, and label

WilliamKelley:
Unfortunately, FormControlLabel does not receive the "error" value of the form control state by default in Mui and apply conditional styling, etc. So, we will start our journey of fixing that here, with accepting it as a prop or from form control context and applying the appropriate classes / logic. The logic is a simple mirror of [Mui's for something like the FormLabel](https://github.com/mui/material-ui/blob/v4.x/packages/material-ui/src/FormLabel/FormLabel.js).


shadoath (approved):
Nice! I was needing this recently

slingner (approved):

PR #438, opened 2022-06-17
feat: use flex gap instead of margin on form control patterns

WilliamKelley:
Components should probably not have outer margin by default, only when a super-component or pattern requires it to. Since I was able to get the designs between Checkbox/Radio group fields and the input/select/text fields to align more, this is realizable for the standard case with a general flex gap.

Note that the fieldset/legend tag patterns are required for checkbox and radio groups, but those come with their own slew of default browser stylings. While fieldset can be reset pretty effectively, legend maintains a weird behavior of some un-resettable, position-absolute-ness that takes it out of the typical flex group pattern. Thus, it must be wrapped in a span in that scenario.


libs/spark/src/Unstable_FormLabel/Unstable_FormLabel.tsx
```diff
@@ -30,15 +30,14 @@ const useStyles = makeStyles<Unstable_FormLabelClassKey>(
...
-      marginBottom: 8,
       /* focused -- can get from internal context => can't condition on prop */
       '&.Mui-focused': {
         // override Mui default
         color: theme.unstable_palette.text.heading,
       },
       /* error -- can get from internal context => can't condition on prop */
       '&.Mui-error': {
-        color: theme.unstable_palette.red[700],
+        color: theme.unstable_palette.text.heading,
```
shadoath, line 40: Why text.heading here?

shadoath (approved):

libs/spark/src/Unstable_FormLabel/Unstable_FormLabel.tsx
WilliamKelley: I'm preferring to use the `palette.text` key instead of a `palette.neutral` key where it is the color of text. Whether or not its idiomatic to use the designated heading key for a form label is a different issue, but it is text. And if you're asking why it's red not neutral, that's the decided design.

shadoath: I was more thinking why not red as it is an error state.

WilliamKelley: Ah gotcha. Yeah the rest of the control will be in error styling: input border, value, sub-control labels. Just not the form label.

shadoath: That makes sense then!

PR #442, opened 2022-06-17
feat(Unstable_FormHelperText): leadingIcon; reserveLineHeight; text list children

WilliamKelley:
Support more complex rendering patterns in FormHelperText

Adds examples of such rendering to individual stories and in fieldset control like RadioGroupField.

Makes the leading icon support more official with own prop. Removes the browser/HTML default styling of excess margin and padding. Instead, only the necessary leading margin for the list item symbols is supplied.


shadoath (approved):
NICE!

slingner (approved):

PR #444, opened 2022-06-21
feat(Unstable_CheckboxGroupField): initial implementation

WilliamKelley:
Similar to RadioGroupField except it employs a more basic `FormGroup` for the controls. We basically aim to achieve [WAI-ARIA's standard on "Two-State" checkbox group](https://www.w3.org/TR/2018/NOTE-wai-aria-practices-1.1-20180726/examples/checkbox/checkbox-1/checkbox-1.html).

Some future enhancements that may be possible:

- Create `CheckboxGroup` that accepts `value` and `onChange` with custom signatures specific to handling the case of multiple choices.
  - In the above, also provide a simple API / handling for a ["Mixed-State" checkbox group](https://www.w3.org/TR/2018/NOTE-wai-aria-practices-1.1-20180726/examples/checkbox/checkbox-2/checkbox-2.html) with indeterminate.
- Add tests to ensure WAI-ARIA requirements are met since we're venturing outside of just extending/copying MUI.

Aside: fixes a story name in RadioGroupField stories


shadoath (approved):
Nice to see this coming together, I can see how we will be using it.

slingner (approved):

PR #514, opened 2022-07-05
feat(Unstable_ListItem): initial implementation

### LINKED ISSUE #516: Migrate ListItem

WilliamKelley:
Base #524 
Closes #516 

This is an un-specified component in Figma, but will be the basis for the Menu Item.

List Item is a very versatile component, representing a series of related content labels or sentences, icon and avatar, and actions.

The API is an extension of the [underlying MUI component](https://v4.mui.com/components/lists/#lists) but with more complexity; it absorbs ListItemIcon, ListItemAvatar, ListItemText, and ListItemSecondaryAction -- none of which are responsible for any significant logic. I figure that no one wants to import all those things that should be baked in already, and they would hurt adoption.

There are a few things that fall outside of Mui's base:
- primary action, `primaryAction`: a separate slot from the leading element (so both can be displayed), meant to contain an action / input that is synonomous with the list item. As such, the entire list item _should act_ as a trigger for the action
    - **note**: this is not ensured by default. it's a non-trivial distribution of props. I plan to create a CheckboxListItem component with a similar intent to CheckboxField: a simple wrapper that ensure an accessible and correctly functioning compound component pattern. You'll notice that the stories are not complex enough to deliver on the purpose that the whole list item should trigger the input, but they are there now for completeness and will be removed when CheckboxListItem is implemented.
- secondary action, `secondaryAction`: uses an underlying Mui component ListItemSecondaryAction because there's logic in the underlying ListItem that detects it. The point is to contain a closely related but separate action / input from the list item. The underlying ListItem will detect its presence and adjust itself so that the action can be triggered without overlapping and also triggering the list item as a button.
- primary and secondary content together: when both are specified, the primary content is automatically boldened. Since the most common design has a label-variant-like top-line and a body-variant bottom-line, this delivers on that by default. However when primary content is alone, it does not need to distinguish itself from secondary content and receives regular body font weight.
- tertiary content, `tertiary`: i predict there will be a design or usefulness in having accompanying text that takes on the description variant which is smaller and lighter. This is a very common design pattern in the wild. It also has the benefit of not triggering the boldened-primary content condition. 
- accessible content levels: the `aria-describedby` attribute is set on the primary content to the ids of the secondary and/or tertiary content, when they are given. you can read [on the MDN page](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-describedby) that this is valid and helpful for any series of tightly related content. 

One example in the wild that i took as inspiration for these features is GitHub's multi-select combobox in the new Projects board interface under the Assignees column. You can see the presence of an checkbox (action), avatar (el), label (primary) and additional info (tertiary) and inspect the DOM and see the aria attributes accordingly.

In addition to the typical prop-validating stories, I added a few descriptively named examples. One of which achieves the above example of GitHub's implementation (aside from the fact it needs to be the full CheckboxListItem pattern to function correctly). Another is a list item from the [Marketplace Microschool details page](https://www.figma.com/file/JgOcyxoJy2Ci8xUbjTQVfU/Microschool-details?node-id=304%3A3711).


WilliamKelley:
This is outdated, will push up new changes soon

libs/spark/src/Unstable_ListItem/Unstable_ListItem.stories.tsx
```diff
@@ -0,0 +1,285 @@
...
+
+export const SecondaryAction: Story = Template.bind({});
+SecondaryAction.args = { secondaryAction: '(IconButton)' };
+SecondaryAction.storyName = 'secondaryAction=(IconButton)';
+
+// ============
+// = Examples =
+// ============
+
+export const SiteInspected: Story = Template.bind({});
```
slingner, line 259: I like these examples, nice addition!

shadoath (approved):

slingner (approved):

PR #515, opened 2022-07-06
feat(Unstable_CheckboxListItem): initial implementation

WilliamKelley:
Based on #514 

Follows through with a simpler component for the checkbox primary action pattern that handles all needs and provides a clean, desirable interface.

misc:
- moves the Checkbox Student example story from List Item to Checkbox List Item.
- adds an example to List Item that emulates "Find a microschool" from the Enrollments app.


libs/spark/src/Unstable_CheckboxListItem/Unstable_CheckboxListItem.tsx
```diff
@@ -0,0 +1,205 @@
...
+    ...other
+  } = props;
+
+  const classes = useStyles();
+
+  const primaryId = useId(primaryTypographyProps?.id);
+  const secondaryId = useId(secondaryTypographyProps?.id);
+  const tertiaryId = useId(tertiaryTypographyProps?.id);
+
+  if (process.env.NODE_ENV !== 'production') {
```
shadoath, line 124: This is new, why add this?

shadoath (approved):

libs/spark/src/Unstable_CheckboxListItem/Unstable_CheckboxListItem.tsx
WilliamKelley: I think it's in some other components. But its to warn consumers in development when they're using it wrong but not in production.

PR #541, opened 2022-08-19
feat(Unstable_ListItem): add nested prop

WilliamKelley:
Very useful when having a primary action of a checkbox and needing to indent the entire element (not just the content group) under an item with an indeterminate grouping


DapperFox (approved):

shadoath (approved):
Very nice, thank you

PR #556, opened 2022-08-24
feat!: remove everything unstable

WilliamKelley:
In preparation for releasing a stable v1 of @prenda/spark, I removed everything unstable. There are no feature changes, just deletions.

Next steps:
- [ ] PR for v1.0.0 release
- [ ] Release v1.0.0
- [ ] Create perma-branch for v1
- [ ] PR for unstable -> stable
	  - [ ] Restore everything unstable (revert this commit)
	  - [ ] Remove everything from v1 that has an unstable replacement.
	  - [ ] Remove "unstable" from everything "unstable".
- [ ] PR for v2.0.0-alpha.1 release
- [ ] Release v2.0.0-alpha.1

Why? 
- We need to reduce our chromatic stories usage. 
- People don't like "unstable" and I presume are scared of using the components as a result.
- We don't need to carry around the cruft of v1 components
- We want to eliminate all need to use hacks to get around v1 styles from the Spark Theme Provider -- we will do so by internalizing the @material-ui/core dependency in a subsequent v2 alpha.


shadoath (approved):
It's gone, its all GONE!

PR #708, opened 2023-04-22
fix: cast theme from MuiTheme to Theme

jb-1980:
### Problem
When using `theme` from `@prenda/spark` in a Remix app, it shows type errors for `theme.palette_alpha` and the like. Investigating the type, the `createTheme` function returns a `MuiTheme` type as seen [here](https://unpkg.com/browse/@prenda/spark@2.0.0-alpha.18/theme/theme.d.ts). 

### Solution
Use type casting to cast it to the `Theme` type, extending the `MuiTheme` with the PDS-specific properties.  
(This could have also used `DefaultTheme` here. Not sure which is preferable.)

### DIFF
--- a/libs/spark/src/theme/theme.ts
+++ b/libs/spark/src/theme/theme.ts
@@ -39,7 +39,7 @@ const theme = createTheme({
   ...initialTheme,
   props,
   overrides: overrides(initialTheme),
-});
+}) as Theme;
 
 theme.unstable_borders = borders_alpha;
 theme.borders_alpha = borders_alpha;


WilliamKelley:
Are you importing the package's module augmentation? I'm guessing not, and that's why you're not seeing those properties in your app. You just need the following line, once, anywhere in your app.
```
import "@prenda/spark/theme/themeAugmentation"
```

In the future, please file issues separately from PR's and make PR's that simply "Close #XXX" and elaborate on the changes made. This would allow me to respond to just the problem and close issues that create history, rather than close PRs.

shadoath:
@WilliamKelley  should the knowledge of this import be in the README of the project?

WilliamKelley:
> @WilliamKelley should the knowledge of this import be in the README of the project?

Yeah

PR #712, opened 2023-11-07
Add MUI v5

shadoath:
Adds packages for MUIv5

### DIFF
--- a/packages/components/package.json
+++ b/packages/components/package.json
@@ -2,6 +2,9 @@
   "name": "@prenda/components",
   "version": "0.0.1",
   "dependencies": {
+    "@mui/base": "^5.0.0-beta.23",
+    "@mui/icons-material": "^5.14.16",
+    "@mui/lab": "^5.0.0-alpha.152",
     "@mui/material": "^5.14.17",
     "tslib": "^2.3.0"
   },
--- a/packages/components/package.json
+++ b/packages/components/package.json
@@ -2,6 +2,8 @@
   "name": "@prenda/components",
   "version": "0.0.1",
   "dependencies": {
+    "@emotion/react": "^11.11.1",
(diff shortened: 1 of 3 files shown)


WilliamKelley:
Let's install things as we use/need them -- so, with an initial component. This is better than just guessing what you need.

We won't want the "icons material" package installed for the components package -- in fact we won't want it at all.

We'll have to look further into whether we keep Emotion as a peer dep, like MUI does, or internalize it. (We definitely want to internalize `@mui/base` or `@mui/material` though)

We should put `react`, `react-dom` as peer deps, that's straightforward enough.

packages/components/package.json
```diff
@@ -2,6 +2,10 @@
   "name": "@prenda/components",
   "version": "0.0.1",
   "dependencies": {
+    "@mui/base": "^5.0.0-beta.23",
+    "@mui/material": "^5.14.17",
+    "react": "^18.2.0",
+    "react-dom": "^18.2.0",
```
WilliamKelley, line 8: These should be peer-deps, not regular deps. 

This `package.json` is what will be published. If we publish this, then consumers have two copies of React. One that they install and import from, and one that we install and import from. No bueno.

For direction, see the [`package.json` of `@mui/material-next`](https://github.com/mui/material-ui/blob/master/packages/mui-material-next/package.json). Ours will look similar.

WilliamKelley:
Ah, I realize you're also making a mistake on how to add dependencies for project development. With a monorepo like this, you will install dependencies at the root of the repository. Nx will take care of recognizing which dependencies the project actually uses when running things in local development. It would also correctly populate the package.json with the relevant dependencies, but I'm not sure how it can be configured to know whether something's a peer/dev dep or not, so I tend to just manually specify the inner package.json

shadoath:
Will restart with new base

PR #724, opened 2023-11-28
chore: add husky pre-commit hook

shadoath:


WilliamKelley:
Couple meta notes
- Please use conventional commit format for your PR titles (since we Squash & Merge, the title becomes the commit name). 
- Please always have a description describing what you did, where it comes from, etc.

Husky is not installed, so this won't work for an isolated system. Are they working for you? If so, probably because you have it installed globally? Please follow Husky's instructions https://typicode.github.io/husky/getting-started.html

Also, please check the husky config when it was removed in #720. We need it to run "lint fix" and "format write". There's a subtle nuance to getting changes staged for commit, a correct adding of husky will ensure this behavior.

shadoath:
Can do this another time

PR #725, opened 2023-11-28
Sb/add theme basics

shadoath:
Initial code for expanding a theme.


shadoath:
@WilliamKelley, what are your thoughts on adding the theme so far? Right direction?

packages/components/src/lib/theme.tsx
```diff
@@ -0,0 +1,39 @@
...
+    brand: true;
+  }
+
+  interface Palette {
+    // this will make the node `secondary` configurable in `extendTheme`
+    // and add `secondary` to the theme's palette.
+    brand: {
+      blue: string;
+      lightBlue: string;
+    };
```
WilliamKelley, line 16: Does this not have to be `brand: PaletteRange`? That's what the [docs](https://mui.com/joy-ui/customization/theme-colors/#adding-new-palettes) would suggest. Can you link in your PR description to whatever you're following? That's the only what i can judge correct-ness.

packages/components/src/lib/theme.tsx
```diff
@@ -0,0 +1,39 @@
...
+      palette: {
+        brand: {
+          blue: '#0A4872',
+          lightBlue: '#D7F3FF',
+        },
+      },
+    },
+  },
+  fontFamily: {
+    body: '"Poppins", sans-serif',
```
WilliamKelley, line 31: Poppins is only used for headings. Normal body text is Inter.

WilliamKelley (commented):
> @WilliamKelley, what are your thoughts on adding the theme so far? Right direction?

I'm not sure, what is "correct" to do to the theme depends on (1) how the configurated values affect components and (2) the structure and properties of the theme that is available to consumers.

Consumers should expect theme object that matches or is compatible with all our design system's properties. Knowing the best way to achieve that while customizing components requires that you have a component developed with our desired Props API and Styles. That's why I'm suggesting you do this only in tandem with a Button.

packages/components/src/lib/theme.tsx
WilliamKelley: Also this seems like it would expose `color="brand"` as an option for color-enabled components. But "brand" is not a valid color for us and is not our desired API. So this would be a wrong way to get these values in there.

shadoath: I tried `PalletteRange` and got the following error when I tried to set a value to `blue`:
```
Type '{ blue: string; lightBlue: string; }' is not assignable to type '{ 50?: string | undefined; 100?: string | undefined; 200?: string | undefined; 300?: string | undefined; 400?: string | undefined; 500?: string | undefined; 600?: string | undefined; 700?: string | undefined; ... 27 more ...; solidDisabledBg?: string | undefined; }
```
Which is why I went with being specific to each one.

packages/components/src/lib/theme.tsx
shadoath: Right, I was putting something there as a placeholder. Where is all of the body/h1 => font documented?
I looked [here](https://www.figma.com/file/w9lodWeivyS4GHdBNIYaJr/foundations?type=design&node-id=2268-228&mode=design&t=qExCpfiVMmWAFkKU-0)

WilliamKelley: There's no defined mapping because there shouldn't be. See my comment in Figma on yours. Variants and the underlying html tag (component) are rightfully independent.

